PR #209, opened 2015-02-18
Rename CertificationAddress into Address in Voice

pschambacher:
@zendesk/mongooses

@ggrossman Since this is public we'll keep the endpoint for a while but we plan to remove it.
There's no documentation in our developer portal for the endpoint and you wouldn't know that you can use this resource unless you dug in the source code. How long do you think that we need to maintain the endpoint?


scott2211:
Looks good to me :+1:

PR #219, opened 2015-03-27
Added support for the new mobile SDK bulk device deletion.

goyox86:
:cloud: :iphone:  

This PR moves the adds support via the 'PushNotification.destroy_many' to the recently created V2 
API entrypoint for unregitering push notification devices (See https://github.com/zendesk/zendesk/pull/16073) by the Mobile SDK team.

NOTE: I tried to emulate what is done for 'Ticket.import'  mainly because this entrypoint is not as 
REST as we could wish (It uses an HTTP POST verb for deletion) and was not a complete CRUD. Any
suggestions are welcome!

/cc @andreionut @steved 

Details:

- Added new PushNotificationDevice resource.
- Added PushNotificationDevice.destroy_many action.
- Added VCR fixtures.
- Added tests.

 ### Steps to merge
* :+1: from the mobile team
* :+1: from the ruby gem team

### References
* https://zendesk.atlassian.net/browse/MSDK-567
### Risks
* None


andreionut:
:+1: from :iphone:

goyox86:
@steved Any feedback on this PR.

P.S: I'm assuming you are the maintainer of the gem.

Regards,

lib/zendesk_api/resources.rb
```diff
@@ -857,4 +857,16 @@ class Variant < ZendeskAPI::Resource
       has_many Variant
     end
   end
+
+  class PushNotificationDevice < Resource
+    post :destroy_many
```
steved: I don't think you need this

lib/zendesk_api/resources.rb
```diff
@@ -857,4 +857,16 @@ class Variant < ZendeskAPI::Resource
...
+
+    def self.resource_path
```
steved: Or this, you can just set a path for the collection

lib/zendesk_api/resources.rb
```diff
@@ -857,4 +857,16 @@ class Variant < ZendeskAPI::Resource
       has_many Variant
     end
   end
+
+  class PushNotificationDevice < Resource
```
steved: Is this actually a resource? Or just a `destroy_many` endpoint?

lib/zendesk_api/resources.rb
```diff
@@ -857,4 +857,16 @@ class Variant < ZendeskAPI::Resource
...
+    post :destroy_many
+
+    def self.resource_path
+      "push_notification_devices/destroy_many"
+    end
+
+    def self.destroy_many(client, tokens)
+     ZendeskAPI::Collection.new(client, self, "push_notification_devices" => tokens, :verb => :post).fetch
```
steved: indent

spec/fixtures/cassettes/push_notification_devices_destroy_many.json
```diff
@@ -0,0 +1 @@
...ort.zd-dev.com/api/v2/push_notification_devices/destroy_many","body":{"encoding":"UTF-8","base64_string":"eyJwdXNoX25vdGlmaWNhdGlvbl9kZXZpY2VzIjpbImZvbyIsImJhciJdfQ==\n"},"headers":{"Accept":["application/json"],"Accept-Encoding":["gzip;q=1.0,deflate;q=0.6,identity;q=0.3"],"User-Agent":["ZendeskAPI API 1.6.1"],"Content-Type":["application/json"]}},"response":{"status":{"code":200,"message":"OK"},"headers":{"Content-Type":["application/json; charset=UTF-8"],"Content-Length":["1"],"Connection":["keep-alive"],"Status":["200 OK"],"X-Zendesk-Api-Version":["v2"],"X-Frame-Options":["SAMEORIGIN"],"X-Zendesk-Origin-Server":["ASSET7006-jnarvaez.local"],"X-Ua-Compatible":["IE=Edge"],"Etag":["W/\"7215ee9c7d9dc229d2921a40e899ec5f\""],"Cache-Control":["max-age=0, private, must-revalidate"],"X-Request-Id":["37471420eda797afa6b57400730acef7"],"X-Runtime":["1.143524"],"X-Zendesk-Request-Id":["9117c96115f3186ff543"],"X-Powered-By":["Phusion Passenger 4.0.45"],"Date":["Fri, 27 Mar 2015 11:32:25 UTC"],"Server":["nginx/1.4.4 + Phusion Passenger 4.0.45"]},"body":{"encoding":"UTF-8","base64_string":"IA==\n"},"http_version":null},"recorded_at":"Fri, 27 Mar 2015 11:32:25 GMT"}],"recorded_with":"VCR 2.9.3"}
```
steved: Any way we can do this test for real? I think this will fail when we run it for CI

lib/zendesk_api/resources.rb
goyox86: What I need to do is a ```destroy_many``` endpoint.

spec/fixtures/cassettes/push_notification_devices_destroy_many.json
goyox86: Hey @steved could you elaborate more about what do you mean. What I did was to point to one of our test servers and run the tests in order to record the VCR cassettes.

lib/zendesk_api/resources.rb
goyox86: @steved I don't see any other callas to .fetch here. Am I doing something worng by calling this fetch here? I think I am :smile:

spec/fixtures/cassettes/push_notification_devices_destroy_many.json
steved: We run the live tests against our staging servers, so if they can be actual tests against our API code that would be great.

lib/zendesk_api/resources.rb
```diff
@@ -857,4 +857,12 @@ class Variant < ZendeskAPI::Resource
       has_many Variant
     end
   end
+
+  class PushNotificationDevice < Resource
+    def self.destroy_many(client, tokens)
+      ZendeskAPI::Collection.new(client, self,"push_notification_devices" => tokens,
+                                 :path =>  "push_notification_devices/destroy_many",
+                                 :verb => :post).fetch
```
steved: Yes, don't call fetch. Collections are lazy-loaded.

Please also indent everything only 2 spaces off the top line.

lib/zendesk_api/resources.rb
steved: I think this should probably just be a `DataResource` class

andreionut:
Jose is in PTO for the next week so I'll finish this PR.

lib/zendesk_api/resources.rb
ggrossman: `DataResource` is the base class, `Resource` inherits from it

`Resource` is a `DataResource` with full CRUD (POST for create, GET for read, PUT for update, DELETE for destroy)

There's also `DeleteResource` but that supports single `DELETE`, so you probably don't want that as the superclass either.

https://github.com/zendesk/zendesk_api_client_rb/blob/master/lib/zendesk_api/resource.rb#L142-175

andreionut: Thanks for the clarification, Gary. `DataResource` is good for now.

andreionut:
@steved Fixed the issues raised.

lib/zendesk_api/resources.rb
```diff
@@ -857,4 +857,14 @@ class Variant < ZendeskAPI::Resource
       has_many Variant
     end
   end
+
+  class PushNotificationDevice < DataResource
+    def self.destroy_many(client, tokens)
+      ZendeskAPI::Collection.new(
+        client, self,"push_notification_devices" => tokens,
+        :path =>  "push_notification_devices/destroy_many",
```
steved: There's some weird spacing on these two lines:
`self,"push..."` and `:path => __ "push..."`

spec/live/push_notification_device_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+          ZendeskAPI::PushNotificationDevice.destroy_many(client, ["foo", "bar"])
+        end
+      end
+    end
+
+    describe "Non-existing devices" do
+      it "silently ignores the devices" do
+        VCR.use_cassette("push_notification_devices_destroy_many") do
+          ZendeskAPI::PushNotificationDevice.destroy_many(client, ["baz"])
+        end
```
steved: These tests don't really test anything, since it just returns a collection. You can remove this entire test if you want, we don't test absolutely every endpoint.

steved:
+1

spec/live/push_notification_device_spec.rb
andreionut: We can test the API contract. If that changes this will break and we will fix it. It's the least we can do.

steved: Right, but this doesn't actually fire a request. It just returns a Collection object.

PR #221, opened 2015-04-01
Capture error messages in 413 responses

jbdietrich:
Since https://github.com/zendesk/zendesk/pull/15158 was deployed, uploads that exceed a plan's maximum file size fail with `413` instead of `422`. Ideally we'd be able to capture the nice, internationalized errors in these responses like we did when they were still returning `422`s.

Alternatively, we could add a new error class specifically for the new response code. I am not sure which option is preferable here.

/cc @steved, @ggrossman


steved:
Fine by me :+1:

ggrossman:
Thanks for catching :+1:

PR #236, opened 2015-06-05
Consider 3XX and 1XX responses invalid (except for 304)

steved:
@vkmita do you mind confirming I'm not crazy and we don't use any of those response codes?


vkmita:
:+1: Fairly positive we don't

cap3618:
Namsuncho
2015. 6. 6. 오전 5:27에 "Victor Kmita" <notifications@github.com>님이 작성:

PR #237, opened 2015-06-10
Bulk actions!

staugaard:
Support for bulk actions.

/cc @zendesk/octo

### DIFF
--- a/lib/zendesk_api/actions.rb
+++ b/lib/zendesk_api/actions.rb
@@ -151,10 +151,15 @@ module ZendeskAPI
 
   module CreateMany
     def create_many!(client, attributes_array)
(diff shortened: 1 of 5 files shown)


steved:
+1

How are you going to handle the job status? I've been wanting to add a small utility class for it, but I've never really had to use it.

staugaard:
Didn't realize that we already had a `ZendeskAPI::JobStatus` resource defined. Now I return instances of those.

steved:
Huh, neither did I. I meant more along the lines of the actual polling for the job completion.

Still +1

staugaard:
Did some small refactoring.

I might follow up with a PR for polling job statuses.

Still +1?

steved:
yep

PR #238, opened 2015-06-11
Introduce reload!

staugaard:
Implements `reload!` on read resources. This is useful for `JobStatus` and possibly other places to.

Most of the diff is just me putting the find methods inside of a ClassMethods module.

/cc @zendesk/octo

### Risks
 - None


steved:
:+1:

PR #242, opened 2015-06-25
destroy_many uses a comma separated list of ids

staugaard:
destroy_many uses a comma separated list of ids

/cc @zendesk/octo @steved @aghassipour 

### Risks
 - None


jespr:
:+1:

Issue #245, opened 2015-07-14
[Zendesk Voice API] Add tickets resources


yoihito:
There is no tickets resource. https://developer.zendesk.com/rest_api/docs/voice-api/voice_integration#creating-voice-tickets

PR #246, opened 2015-07-14
Add zendesk voice ticket resource #245

yoihito:


lib/zendesk_api/resources.rb
```diff
@@ -785,6 +785,10 @@ class Greeting < Resource
     class GreetingCategory < Resource
       namespace "channels/voice"
     end
+
+    class Ticket < Resource
```
steved: Maybe just a CreateResource?

yoihito: Its make sense.

steved:
Thanks! Released as 1.9.6.

PR #254, opened 2015-07-30
Add agent resource and ticket-display feature

yoihito:


.bundle/config
```diff
@@ -0,0 +1,3 @@
+---
+BUNDLE_PATH: vendor/bundle
+BUNDLE_DISABLE_SHARED_GEMS: '1'
```
steved: Probably okay to add this to .gitignore too

lib/zendesk_api/resources.rb
```diff
@@ -796,6 +796,24 @@ class GreetingCategory < Resource
...
+
+    class Agent < ReadResource
+      namespace "channels/voice"
+
+      class Ticket < CreateResource
+        def new_record?
+          true
+        end
+
+        def self.display(client, options)
```
steved: Could this be an instance method on Ticket? i.e. `Ticket.new(....).display!`. I'm also okay with this way, just call it `self.display!`.

steved:
Released in 1.11.3, thanks!

PR #256, opened 2015-08-06
Get RecordInvalid message from description in absence of details

jbdietrich:
I found out that even after https://github.com/zendesk/zendesk_api_client_rb/pull/221 we don't get meaningful error messages for uploads that exceed the maximum size. This is because the structure of an upload's errors hash differs from the typical `RecordInvalid` message. 

This PR adds support for falling back to an error's description if details are missing.

@steved, @ggrossman

### DIFF
--- a/lib/zendesk_api/error.rb
+++ b/lib/zendesk_api/error.rb
@@ -10,11 +10,11 @@ module ZendeskAPI
       def initialize(response)
         @response = response
 
-        if response[:body].is_a?(Hash) && response[:body].key?("details")
-          @errors = response[:body]["details"]
-        else
-          @errors = {}
+        if response[:body].is_a?(Hash)
+          @errors = response[:body]["details"] || response[:body]["description"]
         end
+
+        @errors ||= {}
       end
 
       def to_s
--- a/spec/core/middleware/response/raise_error_spec.rb
+++ b/spec/core/middleware/response/raise_error_spec.rb
@@ -102,7 +102,7 @@ describe ZendeskAPI::Middleware::Response::RaiseError do
       end
 
       context "with a body" do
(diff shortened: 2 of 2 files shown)


lib/zendesk_api/error.rb
```diff
@@ -10,8 +10,8 @@ class RecordInvalid < ClientError
       def initialize(response)
         @response = response
 
-        if response[:body].is_a?(Hash) && response[:body].key?("details")
-          @errors = response[:body]["details"]
+        if response[:body].is_a?(Hash) && response[:body].values_at("details", "description").any?
+          @errors = response[:body]["details"] || response[:body]["description"]
         else
           @errors = {}
         end
```
steved: How about 

```ruby
if response[:body].is_a?(Hash)
 @errors = response[:body]["details"] || response[:body]["description"]
end

@errors ||= {}

```

steved:
We should probably fix the discrepancy in the error message too, right?

steved:
+1

lib/zendesk_api/error.rb
jbdietrich: Sure!

jbdietrich:
We could change the structure of the error if we can avoid a breaking change. Here's how it looks at the moment: `{"error"=>"AttachmentTooLarge", "description"=>"Attachment is too large. Limit 20 megabytes."}`

Another option is to create a dedicated error class in the client for these errors.

steved:
Yeah, even just having both for a while is good enough for me. But that's up to @ggrossman. Merge away, let me know if you need a release.

jbdietrich:
Thanks, a release would be great!

steved:
v1.11.1 (best release number ever)

PR #265, opened 2015-09-04
Raise an error if job status query fails

xuebinglee:
:koala:

Current example of job status query does not throw an error if it fails. This is undesirable since the error does not get addressed.

### Risks
 - None

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,7 @@ until %w{failed completed}.include?(body["status"])
 
   sleep(body["retry_in"])
 end
+raise "Job status query failed. body['message']" if body["status"] == "failed"
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -314,6 +314,7 @@ until %w{failed completed}.include?(body["status"])
 
   sleep(body["retry_in"])
 end
+raise "Job status query failed. body['message']" if body["status"] == "failed"
```
steved: You don't actually interpolate `body['message']`

steved:
Maybe just implement the actual code? :trollface:

xuebinglee:
@steved haha good point. Will do on the next lab day.

PR #266, opened 2015-09-04
Handling bad utf-8 characters on a response

fneves:
If a response contains characters with bad UTF-8 characters the ParseIsoDates middleware blows up when doing the Regex. 
This PR is just to allow the client to handle that case.

/cc @henders @steved @grosser

### DIFF
--- a/lib/zendesk_api/client.rb
+++ b/lib/zendesk_api/client.rb
@@ -13,6 +13,7 @@ require 'zendesk_api/middleware/request/url_based_access_token'
 require 'zendesk_api/middleware/response/callback'
 require 'zendesk_api/middleware/response/deflate'
 require 'zendesk_api/middleware/response/gzip'
+require 'zendesk_api/middleware/response/sanitize_response'
 require 'zendesk_api/middleware/response/parse_iso_dates'
 require 'zendesk_api/middleware/response/parse_json'
 require 'zendesk_api/middleware/response/raise_error'
@@ -142,6 +143,7 @@ module ZendeskAPI
         builder.use ZendeskAPI::Middleware::Response::Logger, config.logger if config.logger
         builder.use ZendeskAPI::Middleware::Response::ParseIsoDates
         builder.use ZendeskAPI::Middleware::Response::ParseJson
+        builder.use ZendeskAPI::Middleware::Response::SanitizeResponse
 
         adapter = config.adapter || Faraday.default_adapter
 
--- a/lib/zendesk_api/middleware/response/sanitize_response.rb
+++ b/lib/zendesk_api/middleware/response/sanitize_response.rb
@@ -1,0 +1,13 @@
+require 'scrub_rb'
+
+module ZendeskAPI
+  module Middleware
+    module Response
+      class SanitizeResponse < Faraday::Response::Middleware
+        def on_complete(env)
+          env[:body].scrub!('')
+        end
+      end
+    end
+  end
+end
--- a/spec/core/middleware/response/sanitize_response_spec.rb
+++ b/spec/core/middleware/response/sanitize_response_spec.rb
@@ -1,0 +1,19 @@
+require 'core/spec_helper'
+
+describe ZendeskAPI::Middleware::Response::SanitizeResponse do
+  def fake_response(data)
+    stub_json_request(:get, %r{blergh}, data)
+    response = client.connection.get('blergh')
+    expect(response.status).to eq(200)
+    response
+  end
+
+  describe 'with bad characters' do
+    let(:response) { fake_response("{\"x\":\"2012-02-01T13:14:15Z\", \"y\":\"\u0315\u0316\u01333\u0270\u022712awesome!\ud83d\udc4d\"}") }
+
+    it 'removes bad characters' do
+      expect(response.body.to_s.valid_encoding?).to be(true)
+      expect(response.body['y'].to_s).to eq("\u0315\u0316\u01333\u0270\u022712awesome!")
+    end
+  end
+end
--- a/zendesk_api.gemspec
+++ b/zendesk_api.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.description = %q{Ruby wrapper for the REST API at http://www.zendesk.com. Documentation at http://developer.zendesk.com.}
   s.license     = 'Apache License Version 2.0'
 
-  s.required_ruby_version     = ">= 1.9.3"
+  s.required_ruby_version     = ">= 1.9.0"
   s.required_rubygems_version = ">= 1.3.6"
 
   s.add_development_dependency "bump"
--- a/zendesk_api.gemspec
+++ b/zendesk_api.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.description = %q{Ruby wrapper for the REST API at http://www.zendesk.com. Documentation at http://developer.zendesk.com.}
   s.license     = 'Apache License Version 2.0'
 
-  s.required_ruby_version     = ">= 1.9.0"
+  s.required_ruby_version     = ">= 1.9.3"
   s.required_rubygems_version = ">= 1.3.6"
 
   s.add_development_dependency "bump"
@@ -30,6 +30,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "inflection"
   s.add_runtime_dependency "multipart-post", "~> 2.0"
   s.add_runtime_dependency "mime-types"
+  s.add_runtime_dependency "scrub_rb", "~> 1.0.1"
 
   s.files              = `git ls-files -x Gemfile.lock`.split("\n") rescue ''
   s.test_files         = `git ls-files -- {test,spec,features}/*`.split("\n")


henders:
We've tested this code as a Monkey Patch in Voyager and works fine!

steved:
Interesting. I thought we'd fixed this server-side. Do we want to return invalid strings anyway? Should they be "scrubbed"?

lib/zendesk_api/middleware/response/parse_iso_dates.rb
```diff
@@ -19,7 +19,9 @@ def parse_dates!(value)
           case value
           when Hash then value.each { |key,element| value[key] = parse_dates!(element) }
           when Array then value.each_with_index { |element, index| value[index] = parse_dates!(element) }
-          when /\A\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z\Z/m then Time.parse(value)
+          when String
+            return Time.parse(value) if value.valid_encoding? && value =~ /\A\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z\Z/m
+            value
```
grosser: otherwise we pass invalid encoded string in ... sounds like this will come back to fail another day ...

how about cleaning it up: `value.encode(Encoding::UTF_8, value.encoding, :invalid => :replace, :undef => :replace)`

fneves: We actually sanitise the data in our project after we get the response. We were not sure however if we should actually do it at the this level or just let the clients handle it. We thought that there might be the possibility that there are characters the clients will be able to handle. If we do that we might be destroying their data.
So, in the case this is always a case of bad data I'm up to replace this with a sanitise method, but as we weren't sure we just fixed the gem to not blow up when we get to this case.

grosser: I don't see a usecase for having/keeping bad characters in a date field ...
if it were a text ... maybe ...
Any way we can fix the original issue of 'something responds with bad
characters' ?

fneves: @grosser we don't know if it is a date field, we just know it is a String. On this case it was on the description of a ticket (a very old ticket).

fneves: If no one really sees any possible case that this data (bad characters) might be of any interest to a client, lets just sanitise the data.
@steved when we fixed this on server side was it due to bad characters sent from a client, or just handling the data bad on server side?
Is there any chance this can be an emoji?

steved: The response itself, since I'm sure there's a bunch of bad data in the database. I couldn't find anything after a quick scan, and it looks like yajl doesn't actually care either when encoding. So maybe it is best that we just clean up all the strings in the client.

fneves:
@steved @grosser Removing the bad characters as soon as the response arrives.

lib/zendesk_api/middleware/response/sanitise_response.rb
```diff
@@ -0,0 +1,13 @@
+module ZendeskAPI
+  module Middleware
+    module Response
+      class SanitiseResponse < Faraday::Response::Middleware
+        def on_complete(env)
+          unless env[:body].strip.empty?
+            env[:body] = env[:body].encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
```
grosser: I think what you want is: `"€foo\xA0".encode(Encoding::UTF_8, invalid: :replace, undef: :replace, replace: '')`

because it keeps more characters and it is the same encoding as (hopefully) everything else we load
```
"€foo\xA0".encode(Encoding::BINARY, invalid: :replace, undef: :replace, replace: '')
=> "foo"
"€foo\xA0".encode(Encoding::UTF_8, invalid: :replace, undef: :replace, replace: '')
=> "€foo"
```

ggrossman: Would it make sense to use `String#scrub` when it is available (Ruby 2.1 and later)?

We had similar issues for automatic credit card redaction and either use `String#scrub` or manually encode, see this code:
https://github.com/zendesk/credit_card_sanitizer/blob/master/lib/credit_card_sanitizer.rb#L135-L147

grosser: Nobody is on 2.0 ... right ? ... hardcode scrub ?

ggrossman: I was basically going off the .travis.yml in this repo which still goes back to 1.9.3... you think we can safely drop support for pre-2.1?

lib/zendesk_api/middleware/response/sanitise_response.rb
```diff
@@ -0,0 +1,13 @@
+module ZendeskAPI
+  module Middleware
+    module Response
+      class SanitiseResponse < Faraday::Response::Middleware
```
ggrossman: I think we should spell this `SanitizeResponse`

According to http://grammarist.com/spelling/sanitise-sanitize/, while "sanitise" has become popular outside of North America, it is not even the traditional British spelling and is not recognized by the Oxford English Dictionary.

lib/zendesk_api/middleware/response/sanitise_response.rb
grosser: we can make the gemspec say >= 2.1 to be safe / have nobody use a newer
version on an old project ... but pushing users to upgrade their ruby is
always good :)

fneves: @grosser I tried that before, but it was not working for 2.0.0. Nonetheless what I had done was wrong since the symbols such as ```€``` were being scrubbed as well.

@ggrossman I tried to use the credit card example but it did not work for JRuby. There was a problem with the null character and JSON.parse.

I ended up with a solution that is similar to the credit_card solution, however instead of replacing with the null character, I basically remove the character. This will avoid having different responses for the MRI and for JRuby

Somehow the tests are hanging in the infinite loop test in this jruby-1.7.19  + jdk 1.7.0_71. It works fine with JDK version 1.8+. Any idea how to solve this?

lib/zendesk_api/middleware/response/sanitize_response.rb
```diff
@@ -0,0 +1,21 @@
+module ZendeskAPI
+  module Middleware
+    module Response
+      class SanitizeResponse < Faraday::Response::Middleware
+        def on_complete(env)
+          to_utf8!(env[:body]) unless env[:body].strip.empty?
```
grosser: force_encoding etc should also work fine on empty strings + (I assume) we rarely expect empty strings ... so no need to optimize for them :)

lib/zendesk_api/middleware/response/sanitize_response.rb
```diff
@@ -0,0 +1,21 @@
...
+        end
+
+        def to_utf8!(str)
+          str.force_encoding(Encoding::UTF_8)
+          if ''.respond_to?(:scrub)
```
grosser: more correct would be: `if str.respond_to?(:scrub!)`

lib/zendesk_api/middleware/response/sanitize_response.rb
```diff
@@ -0,0 +1,21 @@
...
+            str.scrub!('') unless str.valid_encoding?
+          else
+            str.encode!(Encoding::UTF_16, :invalid => :replace, :replace => '')
```
grosser: also need `:undefined => :replace`

grosser: a straight up `encode! Encoding::UTF_8, Encoding::UTF_8` does not work ?

spec/core/middleware/response/sanitise_response_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    stub_json_request(:get, %r{blergh}, data)
+    response = client.connection.get('blergh')
+    expect(response.status).to eq(200)
+    response
+  end
+
+  describe 'with bad characters' do
+    let(:response) { fake_response("{\"x\":\"2012-02-01T13:14:15Z\", \"y\":\"awesome!\ud83d\udc4d\"}") }
+
+    it 'should remove the bad characters' do
```
grosser: fyi shorter version: `it 'removes bad characters'`

spec/core/middleware/response/sanitise_response_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+describe ZendeskAPI::Middleware::Response::SanitizeResponse do
+  def fake_response(data)
+    stub_json_request(:get, %r{blergh}, data)
+    response = client.connection.get('blergh')
+    expect(response.status).to eq(200)
+    response
+  end
+
+  describe 'with bad characters' do
+    let(:response) { fake_response("{\"x\":\"2012-02-01T13:14:15Z\", \"y\":\"awesome!\ud83d\udc4d\"}") }
```
grosser: throwing some actual utf-8 in the test might reveal some bad behavior `ÍÎ˛¸◊Ç˛◊´„`

lib/zendesk_api/client.rb
```diff
@@ -142,6 +143,7 @@ def build_connection
         builder.use ZendeskAPI::Middleware::Response::Logger, config.logger if config.logger
         builder.use ZendeskAPI::Middleware::Response::ParseIsoDates
         builder.use ZendeskAPI::Middleware::Response::ParseJson
+        builder.use ZendeskAPI::Middleware::Response::SanitizeResponse
```
grosser: should be the first so things like parsing/logging bad responses does not blow up ?

lib/zendesk_api/middleware/response/sanitize_response.rb
fneves: Not on ruby 2.0.0

grosser: most of our projects are on 2.1 ... maybe that's good enough ?
otherwise whatever works ...

spec/core/middleware/response/sanitise_response_spec.rb
fneves: Was trying to keep it consistent with all the other specs in this gem. But ok, I guess we will have to start to remove the "should" words someday.

lib/zendesk_api/client.rb
fneves: It is the first thing that runs. I've tested this locally and it seems that it runs in the opposite order we declare the middleware.

grosser: ah ok ... weird ...

lib/zendesk_api/middleware/response/sanitize_response.rb
```diff
@@ -0,0 +1,21 @@
...
+  module Middleware
+    module Response
+      class SanitizeResponse < Faraday::Response::Middleware
+        def on_complete(env)
+          to_utf8!(env[:body])
+        end
+
+        def to_utf8!(str)
+          str.force_encoding(Encoding::UTF_8)
+          if str.respond_to?(:scrub)
```
grosser: to keep these consistent how about `return str if str.valid_encoding?` before the `if`

lib/zendesk_api/middleware/response/sanitize_response.rb
```diff
@@ -0,0 +1,21 @@
...
+            str.scrub!('') unless str.valid_encoding?
```
grosser: ... any reason dropping pre ruby 2.1 support and just scrubbing won't work ?

fneves: If we can just assume that that won't be a problem with some customers that might be using the gem using the old 1.9.3 version, I am more than fine with it.

grosser: ah yeah ... better to keep it to not have any customers complain ...

fneves: JRuby is complaining.... scrub is only supported in jruby9000.

Can we force this as well?

grosser: so scrub is defined and does not work ?

fneves:
Scrub is supported on version 9000 of JRuby.
This means I will have to support the old way of removing the bad characters as well. 
Right now all is working apart from JRuby. Travis version that we are running tests against is JRuby 1.7.19 + jdk1.7, which causes my tests to fail. They work correctly in JRuby 9000, or in 1.7.21 with JDK 8.
Have to investigate more to find why this is happening.

grosser:
don't go too crazy ... maybe skip the test on broken ruby versions

lib/zendesk_api/middleware/response/sanitize_response.rb
steved: Can you just use a shim? [scrub_rb](https://rubygems.org/gems/scrub_rb)?

lib/zendesk_api/middleware/response/sanitize_response.rb
```diff
@@ -0,0 +1,21 @@
+module ZendeskAPI
+  module Middleware
+    module Response
+      class SanitizeResponse < Faraday::Response::Middleware
+        def on_complete(env)
+          to_utf8!(env[:body])
+        end
+
+        def to_utf8!(str)
+          str.force_encoding(Encoding::UTF_8)
```
steved: The less we force the encoding, the better. We should really leave it up to ruby to figure out the default encoding.

fneves:
@steved the shim worked. Thanks !

Comments have been addressed.

ggrossman:
Cool, we can go back to https://github.com/zendesk/credit_card_sanitizer and switch to the shim too (although maybe there we can forget Ruby <2.1)

fneves:
@ggrossman opened a PR on the credit_card_sanitizer as well [PR#23](https://github.com/zendesk/credit_card_sanitizer/pull/23)

grosser:
:+1:

steved:
Thanks! Released as 1.12.0.

PR #284, opened 2015-12-23
fix method_as_class to handle non alphanumeric

idanski:
Hey,

My team had a production bug caused by calling .empty? on a client object.
the client api doesn't handle question marks well so I added this handling.

Also added test. :)


steved:
Thanks!

PR #285, opened 2016-01-04
Add agent Set Group Membership as Default

pchhetri:
https://developer.zendesk.com/rest_api/docs/core/group_memberships#set-m
embership-as-default


steved:
Were you able to actually make a call? I swear there was something weird I ran into

pchhetri:
I tested it the following way in irb: https://gist.github.com/pchhetri/369170123952655404c1

Got lost in writing the rspec test for this.

steved:
WFM :+1:

PR #287, opened 2016-01-09
Use currently documented Incremental Ticket API

c-kII:
Update the Ticket.incremental_export method to use the `/api/v2/incremental/tickets.json` endpoint per https://developer.zendesk.com/rest_api/docs/core/incremental_export#incremental-ticket-export

### DIFF
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -454,7 +454,7 @@ module ZendeskAPI
     # @param [Integer] start_time The start_time parameter
     # @return [Collection] Collection of {Ticket}
     def self.incremental_export(client, start_time)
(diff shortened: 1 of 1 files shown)


ianyamey:
+1

Without this fix, `tickets.incremental_export(1.hour.ago).include(:users)` does not correctly side-load the users resource

apurvis:
we have seen that the `incremental_export` doesn't actually return the `ticket_form_id` field and we have to make an individual load to get that to work; i'm thinking this PR might fix that.

atyndall:
+1 this is still an issue

ggrossman:
Thank you for your contribution, @c-kII , and apologies for the very long wait. We have merged https://github.com/zendesk/zendesk_api_client_rb/pull/381 which also fixes this issue. Closing this PR.

PR #288, opened 2016-01-15
Fix a deprecation warning

sandlerr:
:bread:

deprecation as per https://github.com/ruby/ruby/commit/a106b310e5bbe678022b127b9d32bb0a55708002 looks like it has been happening since 2.0

private methods on Array:
```ruby
irb(main):001:0> [].private_methods
=> [:initialize, :initialize_copy, :DelegateClass, :default_src_encoding, :irb_binding, :sprintf, :format, :Integer, :Float, :String, :Array, :Hash, :catch, :throw, :loop, :block_given?, :Rational, :Complex, :trace_var, :untrace_var, :at_exit, :set_trace_func, :caller, :caller_locations, :test, :fork, :exit, :`, :gem_original_require, :sleep, :respond_to_missing?, :load, :exec, :exit!, :system, :spawn, :syscall, :printf, :open, :putc, :print, :readline, :puts, :p, :gem, :readlines, :rand, :gets, :srand, :proc, :lambda, :abort, :initialize_clone, :initialize_dup, :trap, :require, :require_relative, :autoload, :autoload?, :binding, :local_variables, :warn, :raise, :fail, :global_variables, :__method__, :__callee__, :__dir__, :eval, :iterator?, :method_missing, :singleton_method_added, :singleton_method_removed, :singleton_method_undefined]
```

/cc @zendesk/vegemite @steved 

### Risks
 - Collection methods stop working as expected


steved:
thanks

PR #302, opened 2016-05-02
Remove version constraint on mime-types

bquorning:
Clients using Ruby < 2.0 should themselves set a version restriction so they don't install mime-types >= v3.0.


bquorning:
cc/ @jacobat @grosser

jacobat:
:+1:

grosser:
might as well bump the gem to ruby 2.0+ ... but either way 👍

PR #303, opened 2016-05-02
RuboCop

bquorning:
Should we enable some style checks on this repository?

cc/ @jacobat @grosser


Rakefile
```diff
@@ -34,10 +34,18 @@ if defined?(RSpec)
   task :default => "spec"
 end
 
+require 'rubocop/rake_task'
+desc 'Run RuboCop on the lib directory'
+RuboCop::RakeTask.new(:rubocop) do |task|
+  # task.patterns = %w(Rakefile lib/**/*.rb spec/**/*.rb)
+  task.formatters = %w(progress)
+  task.fail_on_error = true
```
grosser: that's not the default ?

.travis.yml
```diff
@@ -8,4 +8,8 @@ rvm:
   - 2.3.0
   - jruby
 
+script:
```
grosser: if it's part of travis it should be part of rake too ... something like https://github.com/abonas/kubeclient/pull/166

Rakefile
```diff
@@ -34,10 +34,18 @@ if defined?(RSpec)
   task :default => "spec"
 end
 
+require 'rubocop/rake_task'
```
grosser: require to the top ?

grosser:
👍 ideally resolve more of the warnings, but that can be done in followup PRs ...

Rakefile
bquorning: Sure thing

.travis.yml
bquorning: Except the rspec tasks are inside a conditional. Don’t wanna mess with that right now.

Rakefile
bquorning: It sure is. Thanks for noticing.

bquorning:
> ideally resolve more of the warnings, but that can be done in followup PRs ...

Yes, more will follow.

PR #304, opened 2016-05-03
More Rubocop

bquorning:
cc/ @grosser

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,3 +5,31 @@ AllCops:
   Exclude:
     - .git/**/*
     - spec/core/middleware/response/sanitize_response_spec.rb
+
+# Align ends correctly.
+Lint/EndAlignment:
+  AlignWith: variable
+
+# Align the elements of a hash literal if they span more than one line.
+Style/AlignHash:
+  EnforcedLastArgumentHashStyle: ignore_implicit
+
+# Align the parameters of a method call if they span more than one line.
+Style/AlignParameters:
+  EnforcedStyle: with_fixed_indentation
+
+# Checks the indentation of hanging closing parentheses.
+Style/ClosingParenthesisIndentation:
+  Enabled: false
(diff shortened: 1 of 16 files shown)


.rubocop.yml
```diff
@@ -5,3 +5,36 @@ AllCops:
   Exclude:
     - .git/**/*
     - spec/core/middleware/response/sanitize_response_spec.rb
+
+# Align ends correctly.
+Lint/EndAlignment:
+  AlignWith: variable
```
grosser: ❤️

.rubocop.yml
```diff
@@ -5,3 +5,36 @@ AllCops:
...
+# Checks the indentation of hanging closing parentheses.
+Style/ClosingParenthesisIndentation:
+  Enabled: false
+
+# Checks the indentation of the first parameter in a method call.
+Style/FirstParameterIndentation:
+  EnforcedStyle: consistent
+
+Style/HashSyntax:
+  EnforcedStyle: hash_rockets
```
grosser: wut ?

lib/zendesk_api/resources.rb
```diff
@@ -819,7 +821,7 @@ def new_record?
 
         def self.display!(client, options)
           new(client, options).tap do |resource|
-            resource.save!(path: resource.path + '/display')
+            resource.save!(:path => resource.path + '/display')
```
grosser: 😿

grosser:
indent/self etc looks good ... just hash_rockets feels like the wrong direction ... maybe allow both and then slowly transition to 1.9 style ...

grosser:
👍 anyway since it's better then before ...

bquorning:
My mistake, I thought we were testing against 1.8, but of course we're testing only as low as 1.9. I'll disallow hashrockets, probably in another PR.

PR #309, opened 2016-06-24
Support for unified Integer class in Ruby 2.4+

bquorning:
Ruby 2.4 unifies Fixnum and Bignum into Integer: https://bugs.ruby-lang.org/issues/12005


bquorning:
cc/ @grosser

sandlerr:
doesn't it need to be conditional on the ruby version?

grosser:
❤️ 👍

PR #318, opened 2016-12-08
Search needs to include query

sideshowcoder:
To perform successful searches the call to the API needs to include `:query`
while `:external_id` does not seem to work.

### DIFF
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -295,7 +295,7 @@ module ZendeskAPI
 
     # Creates a search collection
     def self.search(client, options = {})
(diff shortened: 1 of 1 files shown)


nogates:
:wave: almost 3 years late 😬 

I think you are right, and we are not using `external_id` for anything there... happy to merge the PR as it's if you want to rebase your branch

nogates:
Oh! thanks @ggrossman ! I had a look at the documentation for the tickets search and I didn't find anything. I forgot that we also use the same resource for searching in organizations and users... I guess we should close this one then, given that we do need the `external_id`?

sideshowcoder:
Thanks for following up, to be honest I don't quite remember my usecase anymore and since `:external_id` is required I'm closing this.

PR #319, opened 2017-01-03
Remove dependency on scrub_rb

bquorning:
For people using Ruby 2.1 or newer, this gem is dead weight. Let's just document that if you use an old version of Ruby, you need to install a gem that implements `scrub!` for you.

### DIFF
--- a/spec/core/spec_helper.rb
+++ b/spec/core/spec_helper.rb
@@ -13,6 +13,7 @@ require 'zendesk_api'
 require 'vcr'
 require 'logger'
 require 'stringio'
+require 'scrub_rb' if RUBY_VERSION < "2.1.0"
 
 begin
   require 'byebug'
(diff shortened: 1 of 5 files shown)


spec/core/spec_helper.rb
```diff
@@ -13,6 +13,7 @@
 require 'vcr'
 require 'logger'
 require 'stringio'
+require 'scrub_rb' if RUBY_VERSION < "2.1.0"
```
fneves: Does this consider JRuby?

bquorning: I believe JRuby calls itself "1.9.3". The JRuby build is green 💚

bquorning: My local JRuby 1.7.22 calls itself "2.0.0". The travis log says
```
$ ruby --version
jruby 1.7.19 (1.9.3p551) 2015-02-05 32f5af0 on Java HotSpot(TM) 64-Bit Server VM 1.7.0_76-b13 +jit [linux-amd64]
```
so that one would probably return "1.9.3".

fneves:
👍

PR #326, opened 2017-03-07
show what exactly went wrong

grosser:
Currently getting random errors like: `ZendeskAPI::Error::NetworkError: the server responded with status 500
` ... would be much nicer to receive: `ZendeskAPI::Error::NetworkError: the server responded with status 500 -- get https://fail.zendesk.com/api/v2/incremental/tickets.json?include=users%2Cgroups%2Corganizations%2Cmetric_sets%2Cdates&start_time=1425754064`

@bquorning @vgrigoruk

### DIFF
--- a/lib/zendesk_api/error.rb
+++ b/lib/zendesk_api/error.rb
@@ -1,14 +1,19 @@
+# tested via spec/core/middleware/response/raise_error_spec.rb
 module ZendeskAPI
   module Error
     class ClientError < Faraday::Error::ClientError
       attr_reader :wrapped_exception
+
+      def to_s
+        "#{super} -- #{response.method} #{response.url}"
+      end
     end
 
     class RecordInvalid < ClientError
-      attr_accessor :response, :errors
+      attr_accessor :errors
 
-      def initialize(response)
-        @response = response
+      def initialize(*)
+        super
 
         if response[:body].is_a?(Hash)
           @errors = response[:body]["details"] || response[:body]["description"]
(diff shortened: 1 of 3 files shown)


lib/zendesk_api/error.rb
```diff
@@ -2,16 +2,20 @@ module ZendeskAPI
...
       attr_accessor :response, :errors
 
-      def initialize(response)
-        @response = response
+      def initialize(*)
+        super
 
-        if response[:body].is_a?(Hash)
-          @errors = response[:body]["details"] || response[:body]["description"]
+        if @response[:body].is_a?(Hash)
```
bquorning: There's still an attr accessor for the response. You don't need to change to using ivars.

grosser: using @response feels more consistent to me ... but don't care ... updated

lib/zendesk_api/error.rb
```diff
@@ -2,13 +2,17 @@ module ZendeskAPI
   module Error
     class ClientError < Faraday::Error::ClientError
       attr_reader :wrapped_exception
+
+      def to_s
+        "#{super} -- #{@response.method} #{@response.url}"
```
bquorning: Actually, `Faraday::Error::ClientError` also defines an `attr_reader :response`. So for consistency, you can just use `response` here as well.

https://github.com/lostisland/faraday/blob/master/lib/faraday/error.rb#L6

grosser:
fixed

bquorning:
:+1:

PR #345, opened 2017-09-02
Fix failure spec in master branch

otoyo:
Fix failure spec.
`to_proc` is different from RSpec master.
https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/matchers/built_in/yield.rb#L38


otoyo:
@grosser Would you be able to review this, please?

grosser:
thx for fixing!

PR #347, opened 2017-09-09
Use newest bundler in travis CI (Fix travis CI)

otoyo:
Use newest bundler reported in bundler/bundler#3559

Failure job
https://travis-ci.org/zendesk/zendesk_api_client_rb/jobs/273512217

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ rvm:
   - 2.3.0
   - jruby
 
+install:
+  - "gem install bundler"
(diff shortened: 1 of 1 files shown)


otoyo:
@grosser Would you merge this also, please?

grosser:
looks good, but how about `before_install` and then don't override how `bundle install` works ?

otoyo:
Indeed it might work.

grosser:
thx!

PR #348, opened 2017-09-19
Add option to raise error when rate limited

tommi-lew:
The motivation of this came from us using Zendesk API as part of background jobs. If we are rate limited, we do not want the job to be hogging the background worker, and instead raise an error and be requeued immediately. This will allow us to handle the error and delay the job for a later time. And also for other jobs to be processed asap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,11 @@ client = ZendeskAPI::Client.new do |config|
   # then retry the request.
   config.retry = true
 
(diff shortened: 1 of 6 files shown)


ggrossman:
Thank you for your contribution @tommi-lew . We regret the long delay on merging it.

tommi-lew:
I just noticed that this is merged! Thanks @ggrossman , better late than never :)

PR #354, opened 2017-11-08
Log response body on 400 range

otoyo:
Zendesk API returns the response status 400 range and the error messages as its body when the request was not successful.
https://developer.zendesk.com/rest_api/docs/core/introduction#400-range

But the response status is only output to the log now.
So, I would like to suggest to log the response body also to understand why the request was failure.

### DIFF
--- a/lib/zendesk_api/middleware/response/logger.rb
+++ b/lib/zendesk_api/middleware/response/logger.rb
@@ -18,7 +18,10 @@ module ZendeskAPI
           @logger.debug dump_debug(env, :request_headers)
 
           @app.call(env).on_complete do |env|
-            @logger.info("Status #{env[:status].to_s}")
+            info = "Status #{env[:status].to_s}"
+            info.concat(" #{env[:body].inspect}") if (400..499).include?(env[:status].to_i)
+
+            @logger.info info
             @logger.debug dump_debug(env, :response_headers)
           end
         end
--- a/lib/zendesk_api/middleware/response/logger.rb
+++ b/lib/zendesk_api/middleware/response/logger.rb
@@ -18,7 +18,7 @@ module ZendeskAPI
           @logger.debug dump_debug(env, :request_headers)
 
           @app.call(env).on_complete do |env|
-            info = "Status #{env[:status].to_s}"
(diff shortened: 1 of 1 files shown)


otoyo:
RVM jruby problem fixes at #355

otoyo:
@grosser would you check this and #355 please?

otoyo:
Test passed! 😃

lib/zendesk_api/middleware/response/logger.rb
```diff
@@ -18,7 +18,10 @@ def call(env)
           @logger.debug dump_debug(env, :request_headers)
 
           @app.call(env).on_complete do |env|
-            @logger.info("Status #{env[:status].to_s}")
+            info = "Status #{env[:status].to_s}"
+            info.concat(" #{env[:body].inspect}") if (400..499).cover?(env[:status].to_i)
```
grosser: should be careful here with logging too big things (crashes logging system / is mostly useless anyway since it's html) ... maybe just do to_s and then the first 1000 characters ?

lib/zendesk_api/middleware/response/logger.rb
```diff
@@ -18,7 +18,10 @@ def call(env)
           @logger.debug dump_debug(env, :request_headers)
 
           @app.call(env).on_complete do |env|
-            @logger.info("Status #{env[:status].to_s}")
+            info = "Status #{env[:status].to_s}"
```
grosser: don't need the `.to_s` here ... it's automatic

otoyo: fixed in db1e458

lib/zendesk_api/middleware/response/logger.rb
otoyo: I did not care that so I think so too.

otoyo: fixed in e434367

grosser:
thx!

grosser:
1.16.0

otoyo:
Thank you for merging!

PR #357, opened 2018-01-17
Don't attempt to sideload with empty HTTP response body

tonyhb:
Sideloading fails with VCR during tests when attempting to iterate
through keys of an empty response body.


tonyhb:
I suppose the greater issue is 304 not modified responses being stored in VCR which occasionally fail - depending on the order and time that the tests were made in.

nogates:
@tonyhb Sorry it took a while to look at this. We have updated our test suite that uses VCR, and we haven't seen any error like that. is there a way you could provide an example of a response that causes this issue?

Issue #308, opened 2016-06-15
Is there a way to hit /users/create_or_update with this gem?


magni-:
https://developer.zendesk.com/rest_api/docs/core/users#create-or-update-user

I'm not seeing anything related to it in the code or in the docs.

``` ruby
client.users.create_or_update(name: 'foo', email: 'bar@baz.com')
# doesn't seem to do anything

ZendeskAPI::User.create_or_update(client, {name: 'foo', email: 'bar@baz.com'})
# NoMethodError: undefined method `create_or_update' for ZendeskAPI::User:Class
```

Am I missing something obvious, or is the functionality not there?

geo1004:
You can do it in this way : 

`ZendeskAPI::User.create(client, email: object_email, name: object_name)`

If you use zendesk_rails and you have initialized the client, you can replace 'client' by : 

`ZendeskRails.client`

magni-:
That clearly doesn't work:
```ruby
client.users.create!(name: 'Duplicate User', email: 'duplicate@user.com')
=> ZendeskAPI::Error::RecordInvalid: ZendeskAPI::Error::RecordInvalid: {"email"=>[{"description"=>"Email: duplicate@email.com is already being used by another user", "error"=>"DuplicateValue"}]}
```

magni-:
Here's how I ended up doing it, for those brought here by Google:
```ruby
response = client.connection.post(
  '/api/v2/users/create_or_update.json', 
  {user: {email: EMAIL, name: NAME}},
)
ZendeskAPI::User.new_from_response(client, response)
```

maxkwallace:
I had to reimplement this in my code as well-- couldn't figure out a way to do it via the API client.

jaxn:
Leave off the `!`

Found this via Google and @geo1004 gave the correct answer.

Derekbrenner:
It would be nice to have the create_or_update call as well.

hkmaly:
Any progress on this?

hkmaly:
Note that I tried the #360 and it doesn't seem to work ; specifically, there should be singular_resource_name instead of resource_name and Collection class also needs to be modified.

josephsiefers:
I agree with @hkmaly - #360 misses the mark. It should be `singular_resource_name` [here](https://github.com/zendesk/zendesk_api_client_rb/blob/73a068a63de384c0c66c446b1a234469928254a5/lib/zendesk_api/actions.rb#L190). 

[...]

PR #360, opened 2018-04-05
Add support for create_or_update for user resource

rkallensee:
This is going to add support for the very convenient [create_or_update](https://developer.zendesk.com/rest_api/docs/core/users#create-or-update-user) on a user resource.

To-Do:
- [ ] Live tests?


ggrossman:
Thank you for your contribution, @rkallensee . Apologies for the very long wait to get it merged.

josephsiefers:
I'm a bit confused - I'm not sure this works. See my comment in https://github.com/zendesk/zendesk_api_client_rb/issues/308

PR #370, opened 2018-10-26
Add support to create or update many users and remove unsupported update many users

michaelherold:
The API for this is slightly different than the API in `CreateMany` and `UpdateMany` because the endpoint doesn't support the ID-based query parameters.

I did not add a live test because it appears that none of the bulk actions have them.


ggrossman:
Thank you for your contribution @michaelherold ! We regret the long delay on merging this change.

PR #382, opened 2019-05-22
Add support for `.find` for attachments

gingermusketeer:
Why: So that attachments can be directly accessed by ID rather than needing the ticket ID and comment ID to access it.


nogates:
Hello @gingermusketeer! thanks for the PR, and so sorry I didn't see it sooner, because I could have used https://github.com/zendesk/zendesk_api_client_rb/pull/382/commits/6bbf2074f6a24f862ce9ca9fa0e6bf3447e7d597 to fix the rubocops errors, which I did in a different PR. 

Re: find attachments by id, I do agree with you that is an operation we should support. Though `ReadResource` base class is not really meant for single resources (it does give you the `find` method, but it does not enforce passing an ID to it).

Let me investigate it a bit further and I will get back to you. thanks

gingermusketeer:
@nogates thanks. Let me know if you want me to make any changes!

nogates:
@gingermusketeer I have been thinking a bit more about this, and I am happy with using `ReadResource` as the base class for attachments. However, I am really concern about adding / changing some functionality here before having a more comprehensive test suite. 

So, I am in the process of reviewing the status of the integration tests, and that should (hopefully) take me about a week or so. By then, I hope I've at least one test to check the current attachment functionality before we change it. 

Thanks!

nogates:
@gingermusketeer I finished adding some integration tests as part of https://github.com/zendesk/zendesk_api_client_rb/pull/392... could you please rebase this on master and fix the merge conflicts? If it passes, I would be more than happy to merge your PR

gingermusketeer:
Many thanks!!

PR #388, opened 2019-08-30
Remove forums resource and start using community topics

nogates:
## Description

Forums resource was removed a long time ago, and all `spec:live` tests are failing because VCR cannot find the forums endpoint, so I guess we can remove it completely from the gem too.

Also, I changed the `topic` resource path to point to `community topics` for the same reason (`live` tests were failing because the endpoint does not exist anymore). Maybe, I should have moved `topic` entirely as well, instead of porting it to HC topics? what do you guys think?


ggrossman:
👏

PR #389, opened 2019-09-03
Remove Gemfile.lock

nogates:
## Description

We shouldn't commit `Gemfile.lock` to the library repo (the Gemfile.lock is reserved for apps)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -49,5 +49,5 @@ rule(/^version:bump:.*/) do |t|
   new_version = version_parts * '.'
   File.open(file, 'w') { |f| f.write(version_file.sub(old_version, new_version)) }
 
-  sh "bundle && git add #{file} Gemfile.lock && git commit -m 'bump version to #{new_version}'"
+  sh "bundle && git add #{file} && git commit -m 'bump version to #{new_version}'"
 end


grosser:
my arguments against this are here https://grosser.it/2015/08/14/check-in-your-gemfile-lock/

nogates:
Interesting. I didn't know bundler has changed the policy about checking `Gemfile.lock` into the version control for gems: https://bundler.io/man/bundle-install.1.html#THE-GEMFILE-LOCK

```
When Bundler first shipped, the Gemfile.lock was included in the .gitignore file included with generated gems. Over time, however, it became clear that this practice forces the pain of broken dependencies onto new contributors, while leaving existing contributors potentially unaware of the problem. Since bundle install is usually the first step towards a contribution, the pain of broken dependencies would discourage new contributors from contributing. As a result, we have revised our guidance for gem authors to now recommend checking in the lock for gems.
```

Which is inline with what you are saying @grosser 

I don't mind adding `Gemfile.lock` to git, I wanted to remove it because the `rake version:bump` task was failing when trying to add it to the repo (since the file was ignored).

grosser:
awesome that they changed their policy :D
the `bump` gem usually takes care of updating the Gemfile.lock with the new version ...

grosser:
... that version task can be removed, use `rake bump:patch`

nogates:
> the `bump` gem usually takes care of updating the Gemfile.lock with the new version ...

usually, but since it's [ignored](https://github.com/zendesk/zendesk_api_client_rb/blob/master/.gitignore#L13), the tasks just fails

```
➜  zendesk_api_client_rb git:(master) ✗ rake version:bump:patch
[...]
The following paths are ignored by one of your .gitignore files:
Gemfile.lock
Use -f if you really want to add them.
rake aborted!
Command failed with status (1): [bundle && git add lib/zendesk_api/version....]
```

grosser:
ah ... that's a bug in bump then :)

grosser:
there is code for that though ...
https://github.com/gregorym/bump/blob/master/lib/bump.rb#L129
so maybe just `bundle update bump`

nogates:
I am gonna close this, since as for now, we can use `rake bump:*` instead of this rake task (which I might remove in a different PR). And at some point, maybe when live tests are properly running, we will need to start including `Gemfile.lock` in the repo ?

PR #392, opened 2019-09-09
Run spec:live tests in travis

nogates:
## Description

We have some RSpec examples that run real requests against a Zendesk instance. These test are quite useful since they allow us to test the actual contract and ensure we haven't broken any endpoint and clients can still use all resources.

However, we never spent the time of setting travis to run these test, and, at the same time, it was complicated for contributors to setup its own Zendesk instance to run the test. All of that have caused that these tests have been broken for some time. Additionally, changes in Zendesk implementations have broken some of the functionalities this library provided (such as topics un support, which was removed a long time ago), which of course, wasn't detected by our unit tests.

So, I've decided to spend some time bringing bringing back the live testing, and, as part of this PR, I've:

- Removed obsolete resources (TopicSubscription, TopicTags, TopicVotes, etc...)
- Moved some of the resources to the HelpCenter API (Topics, Categories)
- Fix application tests broken because ZAF applications are now enforcing version 2.0
- Finally, configured travis to run all spec:live test against a dedicated Zendesk instance. 

I think this should give us a bit more confidence in our testing, which will help us to implement newer features or to accept external contributions.


.travis.yml
```diff
@@ -12,6 +12,12 @@ rvm:
 before_install:
   - "gem install bundler"
 
+jobs:
+  include:
+    - stage: spec:live
+      rvm: 2.3
+      script: bundle exec rake clean_live set_travis_credentials spec:live
```
nogates: I decided not to add the spec:live to the common script and not to run it with all rubies for two main reasons:

- It will cause the build to be way slower
- It will cause a few race conditions since we would then have multiple request running in parallel performing the same operations in the same instance.

I think it's fair to assume we can run only one set of these tests. though maybe it we should use the latests available ruby instead of the latest version we support? thoughts?

ggrossman: I think if we're going to run only one set of live tests, it should be with the latest Ruby version (2.6.4). People on 2.3 will ultimately upgrade to 2.6, but nobody on 2.6 is going back to 2.3.
We should update travis to use 2.6.4 instead of 2.6.1 while we're at it.

ggrossman:
@nogates This is fantastic work! 👍

PR #393, opened 2019-09-11
Add project metadata to the gemspec

orien:
Add [project metadata](https://guides.rubygems.org/specification-reference/#metadata) to the gemspec file. This'll allow people to more easily access the source code, raise issues and read the changelog. These `bug_tracker_uri`, `changelog_uri`, `documentation_uri`, `wiki_uri` and `source_code_uri` links will appear on the rubygems page at https://rubygems.org/gems/zendesk_api after the next release.

### DIFF
--- a/zendesk_api.gemspec
+++ b/zendesk_api.gemspec
@@ -13,6 +13,14 @@ Gem::Specification.new do |s|
   s.description = 'Ruby wrapper for the REST API at https://www.zendesk.com. Documentation at https://developer.zendesk.com.'
   s.license     = 'Apache License Version 2.0'
 
+  s.metadata = {
+    'bug_tracker_uri' => 'https://github.com/zendesk/zendesk_api_client_rb/issues',
+    'changelog_uri' => "https://github.com/zendesk/zendesk_api_client_rb/blob/v#{s.version}/CHANGELOG.md",
+    'documentation_uri' => "https://www.rubydoc.info/gems/zendesk_api/#{s.version}",
+    'source_code_uri' => "https://github.com/zendesk/zendesk_api_client_rb/tree/v#{s.version}",
+    'wiki_uri' => 'https://github.com/zendesk/zendesk_api_client_rb/wiki'
+  }
+
   s.files = Dir.glob('{lib,util}/**/*')
 
   s.required_ruby_version     = ">= 2.3"


zendesk_api.gemspec
```diff
@@ -13,6 +13,14 @@ Gem::Specification.new do |s|
   s.description = 'Ruby wrapper for the REST API at https://www.zendesk.com. Documentation at https://developer.zendesk.com.'
   s.license     = 'Apache License Version 2.0'
 
+  s.metadata = {
+    'bug_tracker_uri' => 'https://github.com/zendesk/zendesk_api_client_rb/issues',
+    'changelog_uri' => 'https://github.com/zendesk/zendesk_api_client_rb/blob/master/CHANGELOG.md',
```
orien: Alternatively, rather than the `master` change log, we could link to the change log at the released tag.

```suggestion
    'changelog_uri' => 'https://github.com/zendesk/zendesk_api_client_rb/blob/v#{s.version}/CHANGELOG.md',
```

zendesk_api.gemspec
```diff
@@ -13,6 +13,14 @@ Gem::Specification.new do |s|
...
+    'documentation_uri' => 'https://zendesk.github.io/zendesk_api_client_rb/',
```
orien: I notice you host your own RDoc. Alternatively, use the automatically generated RDoc at www.rubydoc.info.

```suggestion
    'documentation_uri' => "https://www.rubydoc.info/gems/zendesk_api/#{s.version}",
```

nogates: I think we should use the rubydoc instead of the hosted one, which hasn't been updated for the last versions... what do you think @ggrossman ?

ggrossman: Sounds good!

zendesk_api.gemspec
ggrossman: That is probably far more useful, thanks!

nogates:
@orien, could you please commit the change to the ruby doc and rebase on master? Thanks!

orien:
@nogates: How does that look?

nogates:
> @nogates: How does that look?

It looks great :) Spec live tests are passing. Merging!

PR #394, opened 2019-09-11
Use SVG badges on readme, instead of PNG

orien:
The `.png` badges look pretty terrible on high resolution monitors:

image

How about using `.svg` instead:

image


nogates:
Thanks @orien ! that looks much better!

PR #395, opened 2019-09-12
Set a request default timeout for zendesk client

nakulpathak3:
I noticed there's no timeout after a connection has been established. The network can error in this case. I'm not sure if we have some default on the Faraday side or if something is done under the hood for long running requests so I thought I should open this MR as a consideration for this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ client = ZendeskAPI::Client.new do |config|
   # config.adapter = :patron
 
   # Merged with the default client options hash
-  # config.client_options = {:ssl => {:verify => false}}
+  # config.client_options = {:ssl => {:verify => false}, :request => {:timeout => 15}}
 
   # When getting the error 'hostname does not match the server certificate'
   # use the API at https://yoursubdomain.zendesk.com/api/v2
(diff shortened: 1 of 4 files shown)


nakulpathak3:
They should be fixed^ I'm not sure what that 2.6.4 SPEC_LIVE failure is about

nogates:
Hey @nakulpathak3,

Yeah, rubocop errors are fixed. Thanks for that. Don't worry about the `spec_live` failure (those test are not meant to run for PRs coming from forked repos). I think the PR is good for merging and I think we will merge it soon!

nakulpathak3:
That's good to hear. Thanks @nogates!

ggrossman:
Hi, I think it's fine to add a default timeout. However, I think it should be 60 seconds. No Zendesk API call should exceed 60 seconds because we actually kill HTTP requests on the server side if they exceed that amount of time. But there are expensive but legitimate API calls that may take close to 60 seconds, so we shouldn't have the timeout too low or some customers may encounter it and get confused.

nakulpathak3:
Changed to 60 and specific examples to 30

nogates:
spec live tests are 💚 Merging! Thanks for your contribution @nakulpathak3

Issue #410, opened 2019-12-02
Hard dependency on Faraday 0.9.x


romikoops:
zendesk_api.gemspec file contains this line of code

```ruby
s.add_runtime_dependency "faraday", "~> 0.9"
```

Current Faraday version is `0.17.1` (https://rubygems.org/gems/faraday)

What the motivation to limit the gem within faraday `0.9.x`? I saw there were problems with faraday 0.8, but in that case it must be:

```ruby
s.add_runtime_dependency "faraday", ">= 0.9.0", "< 1.0.0"
```

Actually, currently `zendesk_api` gem is not compatible with `google-cloud-ruby` gem, which is official gem for Rails Active Storage https://edgeguides.rubyonrails.org/active_storage_overview.html#google-cloud-storage-service

nogates:
Merged the PR! thanks for the contribution @romikoops ! Closing this

PR #411, opened 2019-12-02
fix hard dependency on Faraday ~> 0.9.x

romikoops:
#410


nogates:
Thanks for the PR @romikoops!! I think we need to review and upgrade a lot of the gems in the Gemfile.

PR #431, opened 2020-09-19
Add update method example to the README

Wingedillidan:
Not being able to directly add/modify ticket attributes then saving has caused some frustration. This is also a bit counter-intuitive when combined with Rails syntax where `.update` is a direct method call to save modified values without needing to `.save`. See https://support.zendesk.com/hc/en-us/articles/226316187-Adding-a-comment-to-a-ticket-via-the-API for more examples of confusion.


nogates:
Thanks for the contribution @Wingedillidan !

PR #443, opened 2021-02-01
Replace mime-types dependency with mini_mime

techAPJ:
The mime-types gem is used to determine the MIME type of the uploaded file via the filename.

The [mini_mime](https://github.com/discourse/mini_mime) library is an alternative which has a lower memory footprint. [This pull request on the mail library](https://github.com/mikel/mail/pull/1059) has some discussion on the topic.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 gem "jruby-openssl", :platforms => :jruby
-gem "mime-types", "~> 2.99", :platforms => [:ruby_19, :jruby]
+gem "mini_mime"
 gem "rake"
 gem "addressable", "< 2.5.0", :platforms => [:ruby_19, :jruby]
 gem "yard"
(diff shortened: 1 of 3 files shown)


techAPJ:
Looks like the test failure is unrelated.

```
ZendeskAPI::Error::NetworkError:
  Failed to open TCP connection to :443 (Connection refused - connect(2) for nil port 443) -- post https:/api/v2/dynamic_content/items
```

techAPJ:
Friendly ping here. @nogates would you be able to review this PR? We would like to get this PR merged and update Zendesk gem version in [discourse-zendesk-plugin](https://github.com/discourse/discourse-zendesk-plugin).

Let me know if this PR needs more work.

swiknaba:
does using mini_mime also solve the compliance issue? See https://github.com/jellybob/mimemagic/pull/3

If this PR is closed, we should at least fix the version of mime-magic to a version that does not come with the compliance problem.

nogates:
Sorry for the delay in merging this. Will release a new version of the gem shortly. Thanks for the contribution @techAPJ

PR #445, opened 2021-02-25
[VEG-207] Add webhooks resource

token-cjg:
@zendesk/vegemite

We should add a new Webhooks resource to this gem, if we would like to use this api client to fetch such resources while using authenticated sessions.

### DIFF
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -775,6 +775,8 @@ module ZendeskAPI
 
   class Target < Resource; end
 
+  class Webhook < Resource; end
+
   module Voice
     include DataNamespace
 
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -776,6 +776,10 @@ module ZendeskAPI
   class Target < Resource; end
 
   class Webhook < Resource; end
+  class Invocation < Resource; end
+  class Webhook < Resource
+    has_many Invocation
+  end
 
   module Voice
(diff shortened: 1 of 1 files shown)


lib/zendesk_api/resources.rb
```diff
@@ -775,6 +775,12 @@ def self.singular_resource_name
 
   class Target < Resource; end
 
+  class Webhook < Resource; end
```
cl8au, line 778: Do we need this?  I thought this file is required for Classic API and we are going to fetch webhooks from Webhook API instead. I may misunderstand the context here. :)

kwent:
I had love to see this available as well!

lib/zendesk_api/resources.rb
ecoologic: Interesting.

Without tests or any actual use of this code, I'm inclined to close this pull request.

Did you gain a better insight and do you think we should complete this now now? @token-cjg & @cl8au?

token-cjg:
This change is no longer needed.

lib/zendesk_api/resources.rb
token-cjg: Pegasus reached out about this recently, as it is desirable for this functionality to be within the ruby client so that they can action a requirement associated to a joint pending project. cc @zendesk/pegasus

I can look into writing one or two tests for this. Sorry got a bit addled and closed it out; I'll open this again and see if I can massage it into shape this afternoon.

PR #451, opened 2021-05-06
Update README links

nickfz3n:
### Description

The class documentation URL has now been updated to the version that is kept up to date automatically. (See also https://github.com/zendesk/zendesk_api_client_rb/pull/393)

All links in the README have been updated to use HTTPS by default.

Some links that were only redirects to other pages are now linking directly to the final destination.

A broken link for the side-loading docs is now pointing to the correct, unbroken page.


antonivanopoulos (approved):
Nice one 👍

nogates:
Thanks @nickfz3n !

PR #455, opened 2021-06-09
Add option to disable resource cache

karina-thompson:
We want an option to be able to turn off the resource cache for use in tests

### DIFF
--- a/lib/zendesk_api/client.rb
+++ b/lib/zendesk_api/client.rb
@@ -40,6 +40,8 @@ module ZendeskAPI
       method = method.to_s
       options = args.last.is_a?(Hash) ? args.pop : {}
 
+      return ZendeskAPI::Collection.new(self, method_as_class(method), options) unless config.use_resource_cache
+
       @resource_cache[method] ||= { :class => nil, :cache => ZendeskAPI::LRUCache.new }
       if !options.delete(:reload) && (cached = @resource_cache[method][:cache].read(options.hash))
         cached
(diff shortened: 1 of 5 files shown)


lib/zendesk_api/configuration.rb
```diff
@@ -42,8 +42,12 @@ class Configuration
     # @return [ZendeskAPI::LRUCache]
     attr_accessor :cache
 
+    # @ereturn [Boolean] Whether to use resource_cache or not
```
dferdian, line 45: Small typo 

```suggestion
    # @return [Boolean] Whether to use resource_cache or not
```

dferdian (approved):
Thanks!

lib/zendesk_api/client.rb
```diff
@@ -40,6 +40,8 @@ def method_missing(method, *args, &block)
       method = method.to_s
       options = args.last.is_a?(Hash) ? args.pop : {}
 
+      return ZendeskAPI::Collection.new(self, method_as_class(method), options) unless config.use_resource_cache
```
KJTsanaktsidis, line 43: I think this is missing the nil check on `method_as_class` from line 50:

```
raise "Resource for #{method} does not exist" unless @resource_cache[method][:class]
```

unless `ZendeskAPI::Collection#initialize` is also doing a nill-check on the `method_as_class` argument?

KJTsanaktsidis (commented):
Other than that, LGTM, nice.

lib/zendesk_api/client.rb
karina-thompson: Good catch! Will fix

KJTsanaktsidis (approved):

andrewbigger (approved):

nogates:
Thank you all :)

PR #464, opened 2021-08-28
ORCD-1012 Network errors are retried if retry = true

cryptomail:
@haljor 
@grahamcruse 
@alanzoppa 
@zenspider 
@snbloch 

Improves retries when layer 3 errors occur and adds 504 network gateway timeouts.
This should improve resiliency a bit when executing the tests.
Adds test and I tested with consumer libs (doesn't crash/hurt/maim consumers)

### DIFF
--- a/lib/zendesk_api/configuration.rb
+++ b/lib/zendesk_api/configuration.rb
@@ -45,7 +45,11 @@ module ZendeskAPI
     # @return [Boolean] Whether to use resource_cache or not
     attr_accessor :use_resource_cache
 
+    # s pecify the server error codes in which you want a retry to be attempted
     attr_accessor :retry_codes
+
+    # specify if you want a (network layer) exception to elicit a retry
+    attr_accessor :retry_on_exception
 
     def initialize
       @client_options = {}
--- a/lib/zendesk_api/configuration.rb
+++ b/lib/zendesk_api/configuration.rb
@@ -45,6 +45,8 @@ module ZendeskAPI
     # @return [Boolean] Whether to use resource_cache or not
     attr_accessor :use_resource_cache
 
+    attr_accessor :retry_codes
+
     def initialize
       @client_options = {}
       @use_resource_cache = true
--- a/lib/zendesk_api/configuration.rb
+++ b/lib/zendesk_api/configuration.rb
@@ -45,7 +45,7 @@ module ZendeskAPI
     # @return [Boolean] Whether to use resource_cache or not
     attr_accessor :use_resource_cache
 
-    # s pecify the server error codes in which you want a retry to be attempted
+    # specify the server error codes in which you want a retry to be attempted
     attr_accessor :retry_codes
 
     # specify if you want a (network layer) exception to elicit a retry
--- a/lib/zendesk_api/middleware/request/retry.rb
+++ b/lib/zendesk_api/middleware/request/retry.rb
@@ -1,5 +1,4 @@
 require "faraday/middleware"
-
 module ZendeskAPI
   module Middleware
     # @private
@@ -14,15 +13,20 @@ module ZendeskAPI
           super(app)
           @logger = options[:logger]
           @error_codes = options.key?(:retry_codes) && options[:retry_codes] ? options[:retry_codes] : DEFAULT_ERROR_CODES
+          @retry_on_exception = options.key?(:retry_on_exception) && options[:retry_on_exception] ? options[:retry_on_exception] : false
         end
 
         def call(env)
           original_env = env.dup
           exception_happened = false
-          begin
+          if @retry_on_exception
+            begin
+              response = @app.call(env)
+            rescue StandardError => e
+              exception_happened = true
+            end
+          else
             response = @app.call(env)
-          rescue StandardError => e
-            exception_happened = true
           end
 
           if exception_happened || @error_codes.include?(response.env[:status])
--- a/lib/zendesk_api/middleware/request/retry.rb
+++ b/lib/zendesk_api/middleware/request/retry.rb
@@ -8,19 +8,32 @@ module ZendeskAPI
       # @private
       class Retry < Faraday::Middleware
         DEFAULT_RETRY_AFTER = 10
-        ERROR_CODES = [429, 503]
+        DEFAULT_ERROR_CODES = [429, 503]
 
         def initialize(app, options = {})
           super(app)
           @logger = options[:logger]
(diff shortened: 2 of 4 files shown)


cryptomail:
Team, I:

- created unit tests
- tested locally by coercing a consumer to consume the new code
- observed correct behavior.

lib/zendesk_api/middleware/request/retry.rb
```diff
@@ -8,7 +8,7 @@ module Request
       # @private
       class Retry < Faraday::Middleware
         DEFAULT_RETRY_AFTER = 10
-        ERROR_CODES = [429, 503]
+        ERROR_CODES = [429, 503, 504]
```
snbloch, line 11: Consider adding 502 as well?

spec/core/middleware/request/retry_spec.rb
```diff
@@ -7,7 +7,7 @@ def runtime
     Time.now.to_f - start
   end
 
-  [429, 503].each do |error_code|
+  [429, 503, 504].each do |error_code|
```
snbloch, line 10: 502 here too...

spec/core/middleware/request/retry_spec.rb
```diff
@@ -21,29 +21,44 @@ def runtime
...
+      end
     end
+  end
 
-    it "should wait default timeout seconds and then retry request on error" do
-      expect(runtime {
-        expect(client.connection.get("blergh").status).to eq(200)
-      }).to be <= 0.5
-    end
+  [503, 504].each do |error_code|
```
snbloch, line 37: 502 here too...

snbloch (approved):
Performs retries on specific network errors.

lib/zendesk_api/middleware/request/retry.rb
```diff
@@ -8,7 +8,7 @@ module Request
       # @private
       class Retry < Faraday::Middleware
         DEFAULT_RETRY_AFTER = 10
-        ERROR_CODES = [429, 503]
+        ERROR_CODES = [429, (500..599).to_a].flatten
```
grahamcruse, line 11: To make this easy for ourselves and customers, and not change current behaviour, I'd suggest creating a global var for the gem with the original array of error codes to be retried automatically and then we are free to set that and tune it as we like in ZBT/ZAT.

cryptomail: Great suggestion, just pushed new config.

lib/zendesk_api/middleware/request/retry.rb
```diff
@@ -8,19 +8,32 @@ module Request
...
           super(app)
           @logger = options[:logger]
+          @error_codes = options.key?(:retry_codes) && options[:retry_codes] ? options[:retry_codes] : DEFAULT_ERROR_CODES
         end
 
         def call(env)
           original_env = env.dup
-          response = @app.call(env)
+          exception_happened = false
+          begin
```
nogates, line 22: This is somehow, changing the default behaviour of the library. And since this library is used in a lot of external and internal projects, I'd prefer to avoid changing any default logic if we can avoid it. For instance, some applications might be expecting this call to raise an exception, and we are changing it.

Instead, I'd suggest adding a configuration directive (as in this [PR](https://github.com/zendesk/zendesk_api_client_rb/pull/455/files)). Something like 

```
#   in lib/zendesk_api/configuration.rb
config.retry_on_exceptions = false
```

Then, in the actual retry file, you could have something like this (pseudo code)

```
begin
  @app.call(env)
  retry! if @error_codes.include?(response.env[:status])
 rescue StandardError => e
   retry! if config.retry_on_exceptions
```

Of course, you would need to enable this config option in you application.

nogates (commented):
Looking good! However, I am suggesting taking a different approach using config options (just to be safe). 

Alternatively, I could just merge the changes, though this might break some apps.

lib/zendesk_api/middleware/request/retry.rb
cryptomail: will add config, was thinking about that as well as the other configurable items.  Adding now.

lib/zendesk_api/middleware/request/retry.rb
```diff
@@ -8,19 +7,37 @@ module Request
...
+            if exception_happened
+              seconds_left = DEFAULT_RETRY_AFTER.to_i
+              @logger.warn "An exception happened, waiting #{seconds_left} seconds... #{e}" if @logger
+            else
+              seconds_left = (response.env[:response_headers][:retry_after] || DEFAULT_RETRY_AFTER).to_i
+            end
 
-          if ERROR_CODES.include?(response.env[:status])
-            seconds_left = (response.env[:response_headers][:retry_after] || DEFAULT_RETRY_AFTER).to_i
             @logger.warn "You have been rate limited. Retrying in #{seconds_left} seconds..." if @logger
```
haljor, line 41: With the more expansive list of conditions for which we're retrying, this logging is too specific.  Not like we're going to pay attention to it every time anyway.

lib/zendesk_api/configuration.rb
```diff
@@ -45,6 +45,12 @@ class Configuration
     # @return [Boolean] Whether to use resource_cache or not
     attr_accessor :use_resource_cache
 
+    # s pecify the server error codes in which you want a retry to be attempted
```
haljor, line 48: /nit "specify"

cryptomail: damn, you got this!  Thank you!

lib/zendesk_api/middleware/request/retry.rb
cryptomail: fixed!

lib/zendesk_api/middleware/request/retry.rb
cryptomail: I think this was there before (and has assertions it happens!)

haljor (approved):

cryptomail:
@nogates never break apps ^_^ when possible. Added ```config.retry_on_exception = true```

nogates (approved):
Happy to merge this as it's! A couple of things though :) 

* On second thoughts, while 429 error is 99.9% of the times a temporal error that will be auto-resolved in less than a minute, a 503 exception might be not that temporal, and the server might be returning 503 for a while.... I wonder whether exception-based retries should also have a counter
* I think it would be good to refactor the `call` method of this middleware. It's got 7 branches already, plus the rescue block.


But as I said, I happy if you merge it as it's now

cryptomail:
@nogates Great points.  I think I'll take this on with my team as a hackday project ^_^.
Super awesome suggestions!

PR #466, opened 2021-10-12
Refine `https` check for URL

ojab:
1. Disallow multiline url where one of the lines starts with `https`
2. Check for `https://` instead of `https`
3. Allow objects responding to `#to_str` like `Addressable::URI`


nogates:
Thank you for the contribution!

PR #467, opened 2021-11-10
Support installations with hashie 5.0.0

eviltrout:
Hashie 5.0.0 is out now, and the restrictions in the gemspec prevent
this gem from working with applications that are using it.

Discourse would like to upgrade to the latest version of this gem
and we cannot support our zendesk plugin without this patch.


eviltrout:
This depends on us being able to update the Zendesk API to support a newer version of hashie. I've made a PR:

https://github.com/zendesk/zendesk_api_client_rb/pull/467

eviltrout:
@nogates is there a procedure or someone I can ping to get a review on this? It's holding up Discourse from updating hashie and I would love to see it merged.

nogates:
Sorry for the delay on getting this merged. Thank you for your contribution

PR #469, opened 2021-12-17
Add `OrganizationMembership.create_or_update{,!}`

ojab:
Didn't tested because I don't have proper zendesk credentials ._.

### DIFF
--- a/spec/live/organization_spec.rb
+++ b/spec/live/organization_spec.rb
@@ -30,7 +30,7 @@ describe ZendeskAPI::Organization, :delete_after do
         end
       end
 
-      it "updates the existing user" do
+      it "updates the existing organization" do
         VCR.use_cassette("create_or_update_find_existing_organization") do
           expect(client.organizations.find(id: organization.id).name).to eql "Updated!"
         end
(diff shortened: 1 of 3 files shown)


spec/live/organization_spec.rb
```diff
@@ -30,7 +30,7 @@ def valid_attributes
         end
       end
 
-      it "updates the existing user" do
+      it "updates the existing organization" do
```
ojab, line 33: Drive-by typo fix that doesn't warrant a separate PR.

ojab:
…and endpoint is 404 :disappointed: 
Should I create a support ticket for that?

nogates:
> Should I create a support ticket for that?

No! it's all good. PR workflows don't actually have permissions to read repository secretes, unless the PR is created within the repo (which is not the case).

But , I fetched this reference and pushed it to a local branch we use to run spec_live tests, and everything seems to be working [fine](https://github.com/zendesk/zendesk_api_client_rb/runs/4691064130?check_suite_focus=true)! so yeah, merging!

Thanks for the contribution!

PR #470, opened 2022-01-09
Fix Linting and CI test Ruby 3.1

ecoologic:
💣 My first PR in this project.

* Currently, master CI is failing. This solves the problem
* Tiny code refactor for simplicity
* Improved README with local run instructions
* Run tests on Ruby 3.1 too, not on 2.5 (eol)

Maybe needs a bump version?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,6 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - 2.5
           - 2.6
           - 2.7
           - 3.0
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,6 +18,7 @@ jobs:
           - 2.6
           - 2.7
           - 3.0
+          - 3.1
           - jruby-9.1
           - jruby-9.2
     steps:
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Zendesk API Client
 
-![Test](https://github.com/zendesk/zendesk_api_client_rb/workflows/Test/badge.svg)
+[![Test](https://github.com/zendesk/zendesk_api_client_rb/workflows/Test/badge.svg)](https://github.com/zendesk/zendesk_api_client_rb/actions/workflows/main.yml?query=branch%3Amaster)
 [![Gem Version](https://badge.fury.io/rb/zendesk_api.svg)](https://badge.fury.io/rb/zendesk_api)
 [![Code Climate](https://codeclimate.com/github/zendesk/zendesk_api_client_rb.svg)](https://codeclimate.com/github/zendesk/zendesk_api_client_rb)
 
@@ -227,6 +227,7 @@ To facilitate a smaller number of requests and easier manipulation of associated
 For example:
 A `ZendeskAPI::Ticket` is associated with `ZendeskAPI::User` through the `requester_id` field.
 API requests for that ticket return a structure similar to this:
+
 ```json
 "ticket": {
   "id": 1,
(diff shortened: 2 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -18,6 +18,7 @@ jobs:
           - 2.6
           - 2.7
           - 3.0
+          - 3.1
```
ecoologic, line 21: Also 2.5 is now eol, so it _could_ be removed, but I would address that separately.

README.md
```diff
@@ -1,6 +1,6 @@
 # Zendesk API Client
 
-![Test](https://github.com/zendesk/zendesk_api_client_rb/workflows/Test/badge.svg)
+[![Test](https://github.com/zendesk/zendesk_api_client_rb/workflows/Test/badge.svg)](https://github.com/zendesk/zendesk_api_client_rb/actions/workflows/main.yml?query=branch%3Amaster)
```
ecoologic, line 3: See the most recent master build, the whole list actually.

lib/zendesk_api/collection.rb
```diff
@@ -324,22 +324,18 @@ def to_param
       map(&:to_param)
     end
 
-    def has_more_results?(response)
-      return false unless response["meta"].present? && response["results"].present?
-
-      response["meta"]["has_more"] && response["results"].length > 0
+    def more_results?(response)
+      response["meta"].present? && response["results"].present?
```
ecoologic, line 328: I'm quite confident these two versions are equivalent, I don't mind a second opinion.

.github/workflows/main.yml
```diff
@@ -14,10 +14,10 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - 2.5
```
ecoologic, line 17: eol

lib/zendesk_api/collection.rb
```diff
@@ -318,28 +318,23 @@ def to_s
       end
     end
 
-    alias :to_str :to_s
-
```
ecoologic, line 322: This is Ruby:

```ruby
alias new_method old_method
alias_method :new_method, :old_method
```

See https://www.rubyguides.com/2018/11/ruby-alias-keyword/

README.md
```diff
@@ -244,11 +245,9 @@ tickets = client.tickets.include(:users)
 # Or client.tickets(:include => :users)
 # Does *NOT* make a request to the server since it is already loaded
 tickets.first.requester # => #<ZendeskAPI::User id=...>
-```
 
-OR
+# OR
```
tagpro, line 249: Can we do this `###` or `####` because `#` should be for title (I am not sure what the norms are at zendesk yet. 

```suggestion
### OR
```

lib/zendesk_api/collection.rb
tagpro: I am thinking that the name should contain `has` as it is a boolean function. Any particular reason for name change?

tagpro (commented):
I have a couple of comments. I am also wondering how did the last merge to master happen? Are there any linting rules mentioned somewhere that needs to more strict?

README.md
ecoologic: that's now inside a code block, I've done this change because it's more consistent with the other code blocks below.

lib/zendesk_api/collection.rb
ecoologic: this change is due to a Ruby convention, enforced by rubocop. Methods ending in `?` don't need a prefix `is` or `has`.

tagpro (approved):
LGTM

README.md
ecoologic: Very good point about the CI rules @tagpro 💯

PR #476, opened 2022-01-13
Implement Webhooks support in the ruby api client

token-cjg:
### Context

Pegasus are after this functionality to support migration of one or two apps from targets to webhooks

### References

* Old PR: https://github.com/zendesk/zendesk_api_client_rb/pull/445
* Example of using ZendeskAPI to create an object (albeit for Triggers, not Webhooks): https://github.com/zendesk/zendesk_browser_tests/blob/c6178624c3feca0082ef0866be73755f2e2dae2b/spec/shared_examples/email/placeholder_suppression_examples.rb#L10-L33


manted (approved):

ecoologic:
I don't know why the diff is so rich, if you check master, most of the diff is already there. I suspect this is github caching the diff.

```
$ g d master
diff --git a/lib/zendesk_api/resources.rb b/lib/zendesk_api/resources.rb
index 5d01c8a..76b65e1 100644
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -835,6 +835,12 @@ module ZendeskAPI

   class Target < Resource; end

+  class Webhook < Resource; end
+  class Invocation < Resource; end
+  class Webhook < Resource
+    has_many Invocation
+  end
+
   module Voice
     include DataNamespace

diff --git a/spec/live/webhook_spec.rb b/spec/live/webhook_spec.rb
new file mode 100644
index 0000000..0526b6e
--- /dev/null
+++ b/spec/live/webhook_spec.rb
@@ -0,0 +1,16 @@
+require 'core/spec_helper'
+
+describe ZendeskAPI::Webhook, :delete_after do
+  def valid_attributes
+    {
+      name: "Random Hook",
+      endpoint: "https://lvh.me",
+      status: :active,
+      http_method: :get,
+      request_format: :json
+    }
+  end
+
+  it_should_be_creatable
+  it_should_be_deletable
+end
```

lib/zendesk_api/resources.rb
```diff
@@ -835,6 +835,12 @@ def self.singular_resource_name
```
ecoologic, line 835: The changes above are already in master, see my other comment.

spec/live/webhook_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'core/spec_helper'
```
ecoologic, line 1: here are the tests. Because Invocations are always empty when you create a Webhook, and given the meta-programming style of these specs (eg `it_should_be_creatable`) I wasn't able to test the empty invocation array in a timely manner, but I confirmed that it appears to be working fine in the console.

ecoologic (approved):
LGTM

PR #478, opened 2022-01-21
Cleanup and documentation

ecoologic:


README.md
```diff
@@ -6,30 +6,22 @@
...
+
+Please report any bug in the [Github issues page](https://github.com/zendesk/zendesk_api_client_rb/issues).
 
 ## Product Support
 
 This Ruby gem supports the REST API's for Zendesk Support, Zendesk Guide,
 and Zendesk Talk. It does not yet support other Zendesk products such as
 Zendesk Chat, Zendesk Explore, and Zendesk Sell.
 
-## Important Notices
```
ecoologic, line 19: these are not important anymore.

lib/zendesk_api/resources.rb
```diff
@@ -277,83 +277,6 @@ class SatisfactionRating < ReadResource
     has Group
   end
 
-  class Search
```
ecoologic, line 280: These are not resources and deserve a file of their own.

lib/zendesk_api/resources.rb
```diff
@@ -277,83 +277,6 @@ class SatisfactionRating < ReadResource
...
-      unless (%w{query external_id} & options.keys.map(&:to_s)).any?
-        warn "you have not specified a query for this search"
-      end
-
-      ZendeskAPI::Collection.new(client, self, options)
-    end
-
-    # Quack like a Resource
-    # Creates the correct resource class from the result_type passed in
-    def self.new(client, attributes)
```
ecoologic, line 294: This is confusing, usually, you would use `initialize`, but in this case, `.new` is a factory method for other classes.

Issue #468, opened 2021-12-02
Searching help center articles with this API? 


aaronpurt:
Hi,

Hoping someone help. The documentation doesn't go into details on interacting with help center articles. 

On Zendesk API searching help center articles we would be able to call: 
https://example.zendesk.com/api/v2/help_center/articles/search.json?query=test&page=1&per_page=25. 

I have setup the client config with url, login info. 
But, how do we go about searching the help center articles in this ruby gem? The documentation isn't very clear how to interact with help center articles. Have tried numerous ways without success. 

Example: client.articles.fetch!(:query => "test") will return 30 results in an array. And it doesn't seem to be filtering correctly. 
But, the result is missing: count, next_page, previous_page.

tythorsen-drip:
@aaronpurt I'm in the same boat. There doesn't seem to be a clear way to search for Guide/Help Center articles. Nothing I've tried so far results in requests to the endpoint suggested by the API docs: `https://{subdomain}.zendesk.com/api/v2/guide/search?`. (from https://developer.zendesk.com/api-reference/help_center/help-center-api/search/)

Zendesk team, would appreciate a response on this.

ecoologic:
@tythorsen-drip and @aaronpurt - you can try out the new feature pointing to the master branch.

Issue #316, opened 2016-09-20
This gem has terrible documentation


josephecombs:
Every single resource in this gem needs a valid code example of every single CRUD action that can be performed for that resource.  Telling us to go read the core API docs is garbage.

jgillman:
@josephecombs Agreed, though I ended up referring to the documentation here which ended up helping me figure out what kind of data I should be sending with what keys: https://developer.zendesk.com/rest_api/docs/core/tickets

You'll have to translate from the raw curl/json requests to them gem yourself though.

Tonkpils:
@josephecombs the beauty of open source is that you can contribute to it freely. A PR instead of an issue goes a long way...specially with docs.

josephecombs:
@Tonkpils this is a SaaS company valued at THREE BILLION DOLLARS.  They can pay some engineers to maintain this gem properly.  NO ONE should work for them for free.

RemyMaucourt:
I do agree, my company is paying Zendesk every month, I think these fees include correct documentation.

apanzerj:
Just because you are paying Zendesk doesn't mean anything with respect to open source software they publish. You aren't paying for this particular piece of software.

josephecombs:
@apanzerj Sure, if you read the exact terms of the seat license payments that most companies pay zendesk, those payments are not paying for working api clients.  It's more a "spirit of the contract" thing.  The fact that this library is hosted inside their github is a testament to that reasonable assertion.

apanzerj:
Seems like what you got here is a [case of the gimmes](https://www.amazon.com/dp/B00472OBVG). If you think this needs better documentation, make a PR, that's how this works. There is no "contract" except the Apache License this uses.

josephecombs:
A now nearly THREE BILLION DOLLAR company should maintain these tools.  Yes, I do expect a company that is deeply integrated SaaS in many, many firms to maintain clients in major languages.  Your posting basically irrelevant accusations here is a major distraction from the important issue raised.

apanzerj:
I still don't understand why you won't make a PR and help out.

josephecombs:
Fair enough, maybe I am being overly critical and I may make a PR. 

There were a lot of idiosyncrasies, from what I recall. I do not remember the exact reason why I opened this issue in the first place, nor do I work at the company where I used this tool.

apanzerj:
[...]

PR #479, opened 2022-01-21
Issue468 search articles

ecoologic:
https://github.com/zendesk/zendesk_api_client_rb/issues/468

I'm making an effort to start improving documentation on all the code that I touch, as this is one of the most commented and old issues we have (https://github.com/zendesk/zendesk_api_client_rb/issues/316). Please help me catch any possible improvement to the docs I can easily achieve.

cc @zendesk/documentation


lib/zendesk_api/collection.rb
```diff
@@ -253,7 +256,7 @@ def next
       if @options["page"]
         clear_cache
         @options["page"] += 1
-      elsif @query = @next_page
+      elsif (@query = @next_page)
```
ecoologic, line 259: because `=` (assignment) can be confused with `==`, Ruby has a convention of adding parenthesis to let the reader know this was a conscious choice.

Personally, I prefer to avoid using assignments in conditionals.

lib/zendesk_api/search.rb
```diff
@@ -0,0 +1,51 @@
...
+      alias resource_path resource_name
+
+      def model_key
+        "results"
+      end
+    end
+  end
+
+  # This will use cursor pagination by default
+  class SearchExport < Search
```
ecoologic, line 43: Using inheritance for sharing code is a little dirty, but I only started this refactoring because Rubocop complained about the diff.

lib/zendesk_api/collection.rb
```diff
@@ -28,7 +29,7 @@ class Collection
     # @param [String] resource The resource being collected.
     # @param [Hash] options Any additional options to be passed in.
     def initialize(client, resource, options = {})
-      @client, @resource_class, @resource = client, resource, resource.resource_name
+      @client, @resource_class, @resource = client, resource, resource.resource_path
```
ecoologic, line 32: This is the actual change that fixes #468.

```
article.resource_name == "articles"
article.resource_path == "help_center/articles"
```

so the right endpoint is hit.

nogates (approved):
LGTM

README.md
```diff
@@ -6,30 +6,22 @@
...
+This Ruby gem is a generic wrapper around Zendesk's REST API. Follow this README and the [wiki](https://github.com/zendesk/zendesk_api_client_rb/wiki) for how to use it.
 
-Please check out the [wiki](https://github.com/zendesk/zendesk_api_client_rb/wiki), [class documentation](https://www.rubydoc.info/gems/zendesk_api/), and [issues](https://github.com/zendesk/zendesk_api_client_rb/issues) before reporting a bug or asking for help.
+You can interact with all the resources defined in [`resources.rb`](lib/zendesk_api/resources.rb). Basically we have some cleaver code to convert Ruby objects into HTTP requests.
+
+Please refer to our [API documentation](https://developer.zendesk.com) for the specific endpoints and once you understand the mapping between Ruby and the HTTP endpoints you should be able to call any endpoint.
+
+The Yard generated documentation is available in at [RubyDoc](https://www.rubydoc.info/gems/zendesk_api).
+
+Please report any bug in the [Github issues page](https://github.com/zendesk/zendesk_api_client_rb/issues).
```
ecoologic, line 17: I haven't yet looked into the wiki, except that I noticed the same content, and I'm not sure how to unify that (yet).

Doesn't concern me at this stage, I'll get there at some point.

lib/zendesk_api/search.rb
```diff
@@ -0,0 +1,51 @@
+# `zendesk_api` gem root
+module ZendeskAPI
+  # A rich factory that returns a class for your searches
+  class Search
+    # Creates a search collection
+    def self.search(client, options = {})
+      if (options.keys.map(&:to_s) & %w[query external_id]).empty?
+        warn "you have not specified a query for this search"
```
tagpro, line 8: What does this warn do? It is just a log from what I found online.

spec/live/article_spec.rb
```diff
@@ -25,5 +25,16 @@ def valid_attributes
...
+    describe "#search" do
+      before { section_article }
+
+      it "finds the article", :vcr do
+        actual = client.articles.search(query: "What")
+
+        expect(actual.count).to be > 0
+        expect(actual.last.title).to eq("What are these sections and articles doing here?")
+      end
+    end
```
tagpro, line 38: I am guessing this is how to write test for the search factory above. Should there also be tests for the cases like nameError or the missing search query as well? I am not sure how does Zendesk/ruby tackles unit tests tbh.

tagpro (commented):
Looks good. just one comment

lib/zendesk_api/search.rb
ecoologic: If properly configured, I believe it can end up in Datadog too, but I didn't investigate much and usually I tend to a logger instance.

https://www.educative.io/edpresso/what-is-warn-method-in-rubys-kernel-object

PR #480, opened 2022-01-25
Bump v1.35.0

ecoologic:
I will push to artifactory once merged.

### DIFF
--- a/lib/zendesk_api/collection.rb
+++ b/lib/zendesk_api/collection.rb
@@ -320,6 +320,8 @@ module ZendeskAPI
       end
     end
 
+    alias to_str to_s
+
     def to_param
       map(&:to_param)
     end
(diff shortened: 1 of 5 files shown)


lib/zendesk_api/collection.rb
```diff
@@ -320,6 +320,8 @@ def to_s
       end
     end
 
+    alias to_str to_s
+
```
ecoologic, line 324: Used to be 

```
alias :to_str :to_s
```

then I removed it, and now I'm re-introducing it the right way

lib/zendesk_api/resources.rb
```diff
@@ -758,7 +758,6 @@ def self.singular_resource_name
 
   class Target < Resource; end
 
-  class Webhook < Resource; end
   class Invocation < Resource; end
   class Webhook < Resource
```
ecoologic, line 762: One definition is enough

tagpro (approved):
LTM

PR #482, opened 2022-02-01
Gemfile: Lock rspec version, newer breaks CI

ecoologic:
Basically, we don't lock our libraries, and there have been some breaking changes these days, so, for now, I'm locking the gems directly in Gemfile, but I have created a ticket to use a lock file.

Details:

https://zendesk.atlassian.net/browse/RED-911
https://zendesk.atlassian.net/browse/RED-913


tagpro (approved):
ez

Issue #321, opened 2017-02-09
Gem posts comment marked private as public


mmayoNR:
When adding more than one private comment to a ticket, after the first comment is submitted, additional comments that should be public will be private. Steps to reproduce:

1. Initialize Zendesk Client - in the rails console:

```
irb(main):087:0> t = zendesk.tickets.find id: 1976
...logging output redacted
irb(main):088:0> t.update(comment: {public: false, body: 'comment 1' })
...logging output redacted
irb(main):089:0> t.save
...logging output redacted
D, [2017-02-09T13:53:46.603165 #10955] DEBUG -- :   Accept: "application/json"
  Accept-Encoding: "gzip;q=1.0,deflate;q=0.6,identity;q=0.3"
  User-Agent: "ZendeskAPI Ruby 1.10.0"
{:ticket=>#<Hashie::Mash comment=#<Hashie::Mash body="comment 1" public=false>>}
...logging output redacted
irb(main):090:0> t.update(comment: {public: false, body: 'comment 2' })
irb(main):091:0> t.save
...logging output redacted
D, [2017-02-09T13:53:57.730411 #10955] DEBUG -- :   Accept: "application/json"
  Accept-Encoding: "gzip;q=1.0,deflate;q=0.6,identity;q=0.3"
  User-Agent: "ZendeskAPI Ruby 1.10.0"
{:ticket=>#<Hashie::Mash comment=#<Hashie::Mash body="comment 2">>}
```

As you can see, on the second comment submission the `public: false` option is for some reason dropped. 

Complete logs from this console session were submitted to Zendesk Support.

PR #483, opened 2022-02-02
Fix issue #321: Ticket Comments should be sent unchanged

ecoologic:
https://github.com/zendesk/zendesk_api_client_rb/issues/321

Ticket "Comments" are not regular attributes to be sent only when they change, but they should be sent un-altered all the time.

### DIFF
--- a/lib/zendesk_api/resource.rb
+++ b/lib/zendesk_api/resource.rb
@@ -146,10 +146,16 @@ module ZendeskAPI
 
     alias :to_param :attributes
 
+    def attributes_for_save
+      { self.class.singular_resource_name.to_sym => attribute_changes }
+    end
+
     private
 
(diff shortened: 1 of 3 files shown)


lib/zendesk_api/resource.rb
```diff
@@ -146,10 +146,16 @@ def inspect
 
     alias :to_param :attributes
 
+    def attributes_for_save
+      { self.class.singular_resource_name.to_sym => attribute_changes }
+    end
+
```
ecoologic, line 152: I'm making this method public, mainly because that way I can test it without having to stub the HTTP requests. I reckon it's not a bad idea, it might be useful.

tagpro (approved):
LGTM

PR #485, opened 2022-02-04
Fix #294 Get article votes

### LINKED ISSUE #294: Is there a way to pull article votes with the gem?
We currently have some extensions that do this but wondering if it would be possible to just use what's in the gem itself... poking around it looks like the answer is "no" but figured it might still be worth asking just in case i missed something

ecoologic:
Allows retrieving article votes:

```ruby
article = client.articles.find(id: 123)
article.votes.fetch!
```

See: 

* https://github.com/zendesk/zendesk_api_client_rb/issues/294
* https://developer.zendesk.com/api-reference/help_center/help-center-api/votes/

### DIFF
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -203,7 +203,9 @@ module ZendeskAPI
 
     has Category
 
+    class Vote < Resource; end
     class Article < Resource
+      has_many Vote
     end
 
     has_many Article
@@ -215,6 +217,9 @@ module ZendeskAPI
         "help_center/articles"
       end
     end
+
+    class Vote < Resource; end
+    has_many Vote
   end
 
   class TopicSubscription < Resource
--- a/spec/live/article_spec.rb
+++ b/spec/live/article_spec.rb
@@ -1,6 +1,6 @@
 require 'core/spec_helper'
 
-describe ZendeskAPI::Article, :delete_after do
+RSpec.describe ZendeskAPI::Article, :delete_after do
   it "expects article to exist" do
     expect(article).not_to be_nil
   end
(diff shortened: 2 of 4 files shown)


spec/live/article_spec.rb
```diff
@@ -34,6 +34,7 @@ def valid_attributes
 
         expect(actual.count).to be > 0
         expect(actual.last.title).to eq("What are these sections and articles doing here?")
+        expect(actual.last.votes).to eq([])
```
ecoologic, line 37: I couldn't find a way to create an upvote (I _think_ that's an area that's completely overlooked in the gem), so this is the best I can do in terms of testing.

lib/zendesk_api/resources.rb
```diff
@@ -203,7 +203,9 @@ def resource_path
 
     has Category
 
+    class Vote < Resource; end
     class Article < Resource
+      has_many Vote
     end
```
ecoologic, line 209: these are section articles votes, because they're nested under `class Section`.

spec/live/topic_subscription_spec.rb
```diff
@@ -8,16 +8,4 @@
...
-  # is using `subscription` as the model name, instead of `topic_subscription`
-  # which is what the save handle_response method is expecting:
-  # https://github.com/zendesk/zendesk_api_client_rb/blob/master/lib/zendesk_api/actions.rb#L5
-  # we should modify it to use the `model_key` if present, but I don't want to do
-  # that now, when I modifying a lot of tests
-  xit "subscription can be found via the topic" do
-    VCR.use_cassette("find_created_topic_subscription") do
-      expect(topic.subscriptions.find(id: subscription.id)).not_to be nil
-    end
-  end
```
tagpro, line 22: I am a bit confused about why are we removing this.

ecoologic: The x in `xit` stands for "temporarily skipped". This has been skipped for years now, so it's not temporary, it's dead code that doesn't work and just makes noise in the test report.

tagpro: If it is a legit todo today, maybe we can create an issue out of this. I'll leave it to you.

tagpro (approved):
LGTM

PR #495, opened 2022-05-12
Add Translation resource

codealchemy:
Following the [API documentation](https://developer.zendesk.com/api-reference/help_center/help-center-api/translations), this will enable CRUD operations for Article translations.

### DIFF
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -205,8 +205,10 @@ module ZendeskAPI
     has Category
 
     class Vote < Resource; end
+    class Translation < Resource; end
     class Article < Resource
       has_many Vote
+      has_many Translation
     end
 
     has_many Article
@@ -221,6 +223,8 @@ module ZendeskAPI
 
     class Vote < Resource; end
     has_many Vote
+    class Translation < Resource; end
+    has_many Translation
   end
 
   class TopicSubscription < Resource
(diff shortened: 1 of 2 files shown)


lib/zendesk_api/resources.rb
```diff
@@ -221,6 +226,8 @@ def resource_path
 
     class Vote < Resource; end
     has_many Vote
+    class Translation < Resource; end
+    has_many Translation
```
aurelieverrot, line 230: Don't we need to add here:
```ruby
has Section
```

And also same between Article and Category? 🤔  
`Category has many Article` and `Article has Category`

ecoologic (approved):
Approving, so I don't block you, but pending solving the provided feedback. ✅

lib/zendesk_api/resources.rb
codealchemy: If I'm following your suggestion right - that might be a nice enhancement to Article / Section / Category relationships, but I don't think it helps translations (at least not familiar with what would be gained by it, in which case may be better suited for a followup PR).

This change follows other examples of Resources being added (ex. [Vote](https://github.com/zendesk/zendesk_api_client_rb/commit/8665b8e9bfb638805be550ab54aca0ac59b86271) or [Plan](https://github.com/zendesk/zendesk_api_client_rb/commit/ec06036f848fbb8fd0847d7d8c8158ea20ff3a37)).

aurelieverrot: I wasn't sure if the relationships had to be "mutual" or not 😊  Seems like it's not the case. Thanks!

aurelieverrot (approved):

PR #496, opened 2022-05-16
Fix faraday deprecation errors for `authorization`

arashm:
Fixes the deprecation errors:

```
WARNING: `Faraday::Connection#authorization` is deprecated; it will be removed in version 2.0.
While initializing your connection, use `#request(:authorization, ...)` instead.
```
I couldn't directly use the client in the test since faraday only edits the headers somewhere using the middleware. So I had to make a request and check the request_header there.

### DIFF
--- a/lib/zendesk_api/client.rb
+++ b/lib/zendesk_api/client.rb
@@ -165,7 +165,7 @@ module ZendeskAPI
 
         # request
         if config.access_token && !config.url_based_access_token
-          builder.authorization("Bearer", config.access_token)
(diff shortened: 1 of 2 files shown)


rsl:
is this gem under active development? this is the second thread/PR/issue about faraday with just zero response from anyone official.

codealchemy (approved):
👍

ecoologic:
FYI: We are working on releasing v2 of the gem that will support Faraday v2.

ecoologic:
You're right@arashm, it would be better. I'll see with the team. 👍

PR #503, opened 2022-08-23
Improve Error Message for Record Invalid Errors

conradbeach:
While attempting to merge 2 users together in Zendesk, I received a 422 response with the response body `{"error"=>"You cannot merge Customer Name while they have an external ID."}`. However, this error message was not included in the raised exception. As a result, our exception monitoring tool only showed `ZendeskAPI::Error::RecordInvalid: {}` without any details. This made determining the actual issue challenging.

I think modifying the code that generates the message for `ZendeskAPI::Error::RecordInvalid` errors to use the message in the `error` key of the response body, if necessary, will improve the usability of this exception.

### DIFF
--- a/lib/zendesk_api/error.rb
+++ b/lib/zendesk_api/error.rb
@@ -20,7 +20,7 @@ module ZendeskAPI
         super
 
         if response[:body].is_a?(Hash)
-          @errors = response[:body]["details"] || generate_error_msg(response[:body])
+          @errors = response[:body]["details"] || generate_error_msg(response[:body]) || response[:body]["error"]
         end
 
         @errors ||= {}
--- a/spec/core/middleware/response/raise_error_spec.rb
+++ b/spec/core/middleware/response/raise_error_spec.rb
@@ -94,6 +94,21 @@ describe ZendeskAPI::Middleware::Response::RaiseError do
             fail # didn't raise an error
           end
         end
+
+        context "with only an error key" do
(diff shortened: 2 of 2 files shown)


lib/zendesk_api/error.rb
```diff
@@ -20,7 +20,7 @@ def initialize(*)
         super
 
         if response[:body].is_a?(Hash)
-          @errors = response[:body]["details"] || generate_error_msg(response[:body])
+          @errors = response[:body]["details"] || generate_error_msg(response[:body]) || response[:body]["error"]
```
codealchemy, line 23: There look to be a few different patterns for errors that could come up in addition to this (eg. `errors` is the key returned in a validation message when attempting to create a translation for a locale that is not supported)

Since this all is returned in the `response[:body]`, it feels like a better fit to consolidate all error message handling in `generate_error_msg` instead of doing some bits here, some there (but not blocking, possibly something to consider - or save for a followup).

codealchemy (approved):
`live` failures in CI appear unrelated (and are failing on `master` as well)

codealchemy:
Included in https://github.com/zendesk/zendesk_api_client_rb/commit/2bd2fd22b126c374f9bccd07b95634c834dac31d, thanks for the patch (and appreciate the patience on this) @conradbeach !

conradbeach:
Thanks, @codealchemy!

Issue #488, opened 2022-02-21
Support Faraday 2.x


tmimura39:
[Faraday 2.x has been released.](https://github.com/lostisland/faraday/blob/fc112254746ae92e3dcc8e6f9055878b1c583cc7/CHANGELOG.md#200-2022-01-04)

but, faraday is require to `< 2.0.0` in the gemspec of zendesk_api.
https://github.com/zendesk/zendesk_api_client_rb/blob/8665b8e9bfb638805be550ab54aca0ac59b86271/zendesk_api.gemspec#L29

Is it possible to relax the dependency?

jrmhaig:
There are breaking changes so simply relaxing the dependency is not enough.

* The `dependency` has been removed. This is used in `lib/zendesk_api/middleware/request/encode_json.rb`.
* `Faraday::Response::Middleware` has been removed. This is used in various places.
* Possibly various other changes too. See the [upgrading guide.](https://github.com/lostisland/faraday/blob/main/UPGRADING.md)

rsl:
so is the plan to just never upgrade? that really hurts those developers who want to keep their software up-to-date. that reply seems a little "not our problem/not planning to address it/old versions work for us".

jrmhaig:
@rsl I absolutely agree and yes, my reply was not an official reply. I was hoping that I had provided enough information so that the Zendesk developers could work on it and am a bit disappointed to not even have received an acknowledgement. I did start creating a fix myself but when I saw that it was more than a few minutes, and we are paying Zendesk for their service, I would really want paying for my time!

We are using Faraday elsewhere in our app and I would like to upgrade those parts to version 2 so this issue is a blocker for me.

rsl:
yeh sadly, we're/i'm looking at ditching this gem [i've ran into other issues besides this] and making our own light wrapper around their REST API. i don't expect them to look into this. i forked the gem and looked at making the changes to make a PR [like a good OSS citizen] but man, those changes are VERY out of my league in terms of knowing how to change out the middleware approaches and things.

i couldn't even get the specs to run out of the box with that json change about `dependency`.

ecoologic:
Hi there,

This update is not a trivial change, but we are likely to pick this up soon. We could surely have done a better job at keeping you informed and we hope to do a better job from now on.

rsl:
[...]

PR #504, opened 2022-08-24
Investigation: Support for Faraday 2

ecoologic:
A first look into issue #488, throw-away investigation.


ecoologic:
This was just an investigation. A separate PR that is working better here: https://github.com/zendesk/zendesk_api_client_rb/issues/488

PR #511, opened 2022-12-15
Add `Article::Label` resource

codealchemy:
cc: @zendesk/i18n 

Per the [API docs](https://developer.zendesk.com/api-reference/help_center/help-center-api/article_labels/), we are able to list / add / remove labels from articles. This adds support for those endpoints to this client.

📝 This can currently be done using `article#update(label_names:)`, which will _overwrite_ the labels on a given article with whatever labels are passed to the update.

### DIFF
--- a/spec/live/article_spec.rb
+++ b/spec/live/article_spec.rb
@@ -11,6 +11,12 @@ RSpec.describe ZendeskAPI::Article, :delete_after do
     expect(article.translations.map(&:locale)).to include("fr")
   end
 
+  it "can have labels", :vcr do
+    article.labels.create(name: "test-label")
+
+    expect(article.labels.map(&:name)).to include("test-label")
+  end
+
   describe "creating articles within a section" do
     def valid_attributes
       { :name => "My Article", user_segment_id: nil, permission_group_id: 2801272, title: "My super article" }
--- a/spec/live/article_spec.rb
+++ b/spec/live/article_spec.rb
@@ -12,9 +12,11 @@ RSpec.describe ZendeskAPI::Article, :delete_after do
   end
 
   it "can have labels", :vcr do
-    article.labels.create(name: "test-label")
+    label = article.labels.create!(name: "ruby-client-test-label")
 
-    expect(article.labels.map(&:name)).to include("test-label")
+    expect(article.labels.map(&:name)).to include(label.name)
+
+    label.destroy
   end
 
   describe "creating articles within a section" do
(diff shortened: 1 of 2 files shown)


spec/live/article_spec.rb
```diff
@@ -11,6 +11,12 @@
     expect(article.translations.map(&:locale)).to include("fr")
   end
 
+  it "can have labels", :vcr do
+    article.labels.create(name: "test-label")
+
+    expect(article.labels.map(&:name)).to include("test-label")
```
ecoologic, line 17: Can we cleanup and delete the label after the test?

codealchemy: > we could make the label name time related, eg "test-label-#{Time.to_i}" to avoid any possible conflict with existing labels.

We could, and this should work for live specs here, but since these tests use VCR (and we don't have Timecop as a dependency... not sure it'd be worth adding here yet?) each subsequent run that uses the cassettes would fail. Switching to using `create!` instead should error early if there is another label with the same name.

 > right now, the latest master is a beta version, so I'm not sure what's the best strategy to release this.
 
It looks like it's been a release candidate since September, so maybe we could cut the next version and add this to it? Or this could be backported to a prior minor version as a patch.

aurelieverrot (approved):

damiann (approved):

spec/live/article_spec.rb
ecoologic: VCR: I haven't looked much into how VCR is setup, but if the cassette is stored in GitHub... how is it reused? I don't think timecop is necessary at this point.  Ultimately, any solution that cleans after itself is fine by me.

Release: I would prefer to release the new version with the exact code of the beta version, after that, a new release won't be that hard.

ecoologic (approved):

spec/live/article_spec.rb
```diff
@@ -11,6 +11,14 @@
     expect(article.translations.map(&:locale)).to include("fr")
   end
 
+  it "can have labels", :vcr do
+    label = article.labels.create!(name: "ruby-client-test-label")
+
+    expect(article.labels.map(&:name)).to include(label.name)
+
+    label.destroy
```
ecoologic, line 19: Ideally, this would be part of the `ensure` block

spec/live/article_spec.rb
ecoologic: Also, this line: https://github.com/zendesk/zendesk_api_client_rb/actions/runs/3833571414/jobs/6525157576#step:4:70

Issue #420, opened 2020-01-28
ticket.safe_update is not respected when updating ticket.tags


justinlang:
Steps to reproduce:

1. Create a ticket
2. Fetch the ticket via the Ruby client in irb
3. Update tags: `ticket.tags = ["tag_via_gem"]`
4. Set `safe_update` and `updated_stamp`:
```
ticket.safe_update = true
ticket.ticket.updated_stamp = DateTime.now.utc.iso8601(0)
```
5. Modify the ticket elsewhere (e.g. in the web UI) by adding a tag, say "tag_via_ui"
6. Note that the ticket in the web UI has saved and now has the tag "tag_via_ui"
6. Do `ticket.save!` in irb

Result: error is returned by the Ruby client: `ZendeskAPI::Error::NetworkError: the server responded with status 409` as expected.
**However**, note that the tags on the ticket are only "tag_via_gem" - the tag change was persisted despite `safe_update` being set to true.

As a user of the gem, I would expect that no changes to the ticket would be made if `safe_update` is set to true.

Examining requests from the gem, I noticed that the gem performs a POST to `/api/v2/tickets/{id}/tags` in order to update tags.  Looking at the docs for this API, it seems like there should be `safe_update` fields added to the request body: https://developer.zendesk.com/rest_api/docs/support/tags#safe-update
But the request body for the POST in the example above just contains:
```
{
  "tags": ["tag_via_gem"]
}
```
This seems like an issue with the way the gem handles tag updates - it's not sending along the `safe_update` fields in the POST to `/api/v2/tickets/{id}/tags`.

justinlang:
Not sure if this is actually supported by the gem:
```
ticket.tags.safe_update = true
ticket.tags.updated_stamp = DateTime.now.iso8601(0)
```
But I can confirm doing this doesn't affect the behaviour I'm seeing.

justinlang:
Verified that this is a bug with the gem.  Used the REST API directly and did PUTs to `/api/v2/tickets/{id}.json` and `/api/v2/tickets/{id}/tags.json` with `safe_update` and `updated_stamp` in the request body.  In both cases tags were not updated if `updated_stamp` didn't match the ticket's `updated_at`, as expected.

datibbaw:
This is primarily a workaround for your issue, but until a better fix is made available, this is what you could use instead
```ruby
ticket.safe_update = true
ticket.updated_stamp = DateTime.now.utc.iso8601(0)
[...]

PR #518, opened 2023-01-22
Red 1545 fix safeupdate for tag

JinHuangAtZen:
Attempt to fix https://github.com/zendesk/zendesk_api_client_rb/issues/420


ecoologic:
This issue will be closed with no code change. The correct solution is to save the ticket as per [this comment](https://github.com/zendesk/zendesk_api_client_rb/issues/420#issuecomment-1333029230).

PS: The solution in this PR was unlikely to solve the problem anyway, see PR521.

PR #521, opened 2023-01-25
RED-1545 Ticket tags safe update

### LINKED ISSUE #420: ticket.safe_update is not respected when updating ticket.tags

ecoologic:
Fix #420.

Associations saved after the main resource.

Note: Update changelog when bumping the gem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ spec/fixtures/cassettes
 spec/fixtures/credentials.yml
 coverage/**
 .yardoc/**
+.ruby-version
 Gemfile.lock
 vendor/bundle/**
 pkg/**
--- a/Gemfile
+++ b/Gemfile
@@ -4,11 +4,8 @@ gem "jruby-openssl", platforms: :jruby
 gem "mini_mime"
 gem "rake"
 gem "addressable", ">= 2.8.0"
-gem "yard"
 gem "json", ">= 2.3.0", platforms: :ruby
 gem "scrub_rb"
-
-gem "rubocop", "~> 0.65.0", require: false
 
 group :test do
   gem "webmock"
@@ -29,6 +26,8 @@ end
 
 group :dev do
   gem "bump"
+  gem "yard"
+  gem "rubocop", "~> 0.65.0", require: false
 end
 
 group :dev, :test do
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem "yard"
 gem "json", ">= 2.3.0", platforms: :ruby
 gem "scrub_rb"
 
-gem "rubocop", "~> 0.64.0", require: false
+gem "rubocop", "~> 0.65.0", require: false
 
 group :test do
   gem "webmock"
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,12 @@ group :dev do
 end
 
 group :dev, :test do
-  gem "byebug", platforms: :ruby
+  platform :ruby do
+    gem "pry"
+    gem "pry-nav"
+    gem "pry-stack_explorer"
+    gem "pry-doc"
+  end
 end
 
 gemspec
--- a/lib/zendesk_api/actions.rb
+++ b/lib/zendesk_api/actions.rb
@@ -27,13 +27,13 @@ module ZendeskAPI
 
       req_path = options[:path] if options[:path]
 
-      save_associations
-
       @response = @client.connection.send(method, req_path) do |req|
         req.body = attributes_for_save.merge(@global_params)
 
         yield req if block_given?
       end
+
+      save_associations
 
       handle_response(@response)
 
--- a/lib/zendesk_api/resource_tag.rb
+++ b/lib/zendesk_api/resource_tag.rb
@@ -4,8 +4,8 @@ module ZendeskAPI
     include Update
     include Destroy
 
-    alias :name :id
-    alias :to_param :id
+    alias name id
+    alias to_param id
 
     def path(opts = {})
       raise "tags must have parent resource" unless association.options.parent
@@ -23,6 +23,7 @@ module ZendeskAPI
       end
     end
 
+    # Overriding
     module Update
       def _save(method = :save)
         return self unless @resources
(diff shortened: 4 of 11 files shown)


spec/macros/resource_macros.rb
```diff
@@ -134,6 +134,8 @@ def it_should_be_readable(*args)
       end if create
 
       after(:all) do
+        return unless @object
```
ecoologic, line 137: This keeps the CI log clean if the resource couldn't be created.

lib/zendesk_api/resource_tag.rb
```diff
@@ -0,0 +1,47 @@
+module ZendeskAPI
```
ecoologic, line 1: todo: delete, this extraction is not loaded.

.gitignore
```diff
@@ -3,6 +3,7 @@ spec/fixtures/cassettes
 spec/fixtures/credentials.yml
 coverage/**
 .yardoc/**
+.ruby-version
```
ecoologic, line 6: This file was not included and we can't include it in a gem, because it needs to be able to work in different versions.

Gemfile
```diff
@@ -29,10 +26,17 @@ end
...
 end
 
 group :dev, :test do
-  gem "byebug", platforms: :ruby
+  platform :ruby do
+    gem "pry"
+    gem "pry-nav"
+    gem "pry-stack_explorer"
+    gem "pry-doc"
+  end
```
ecoologic, line 39: pry > byebug

lib/zendesk_api/actions.rb
```diff
@@ -27,14 +27,14 @@ def save!(options = {})
...
-      save_associations
-
       @response = @client.connection.send(method, req_path) do |req|
         req.body = attributes_for_save.merge(@global_params)
 
         yield req if block_given?
       end
 
+      save_associations
+
```
ecoologic, line 37: ATTENTION! This is the change for this PR:

From the issue 420 (not linking it until I have a proper solution):

> However, note that the tags on the ticket are only "tag_via_gem" - the tag change was persisted despite safe_update being set to true.

This happens because we were saving the associations before the resource, and we don't have transactions on APIs.

lib/zendesk_api/resources.rb
```diff
@@ -40,50 +41,7 @@ class Session < ReadResource
...
-          raise e
-        end
-      end
-    end
-
-    def attributes_for_save
-      { self.class.resource_name => [id] }
-    end
-  end
+  require_relative "resource_tag"
```
ecoologic, line 44: Of course I don't like `require_relative` in the middle of a page, but sorting out all the circular dependencies started to take too long.

lib/zendesk_api/resources.rb
```diff
@@ -367,114 +325,6 @@ def self.incremental_export(client, start_time)
...
-    # @param [Client] client The {Client} object to be used
-    # @param [Hash] attributes The attributes to create.
-    # @return [Ticket] Created object or nil
-    def self.import(client, attributes)
-      ticket = new(client, attributes)
-      return unless ticket.save(:path => "imports/tickets")
-      ticket
-    end
-  end
-
```
ecoologic, line 477: I get lost when files pierce through the 200th line.

Gemfile
Kalimaha: I disagree! 🙂 All good though, your choice 🙂

lib/zendesk_api/resource_ticket.rb
```diff
@@ -0,0 +1,116 @@
+module ZendeskAPI
+  # https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/
+  class Ticket < Resource
```
Kalimaha, line 3: Do we need tests for these new classes?

Kalimaha (commented):
Few minor comments, keen to wait for David feedback.

lib/zendesk_api/resource_ticket.rb
ecoologic: this is an extraction, the tests are already in place.

Gemfile
ecoologic: Interesting. I'd be happy to dig deeper if you have the time, I might learn something.

lib/zendesk_api/actions.rb
ecoologic: Rather than a global option, I was considering a "per request" option, like the above `ticket.save(save_associations: :after)`. That would allow users to fix their problem on the specific call, and not having to review the whole code base.

Solving the second problem using a PUT opens the new one: using PUT instead of POST results in a union of the tags. This would change the behaviour for all our clients. Adding `safe_update` on POST seems like another workaround to not having transactional APIs. Other options are discussed in the card RED-1545.

I'd like to solve the specific problem in the Z2 PT, and leave this second point that no-one yet complained about for another day.

ecoologic:
This issue will be closed with no code change. The correct solution is to save the ticket as per [this comment](https://github.com/zendesk/zendesk_api_client_rb/issues/420#issuecomment-1333029230).

PR #523, opened 2023-02-05
v1.38.0.rc2 Drop support for Ruby 2.6

ecoologic:
See RED-1558.

Message from PO:

It is possible that the SDK will work okay in Ruby 2.6, but we do not guarantee it and also support will not be provided on any issues.

We will continue to support Ruby 2.7 owing to the large set of developers still on Ruby 2. The support will end by Jan 2024.

We strongly recommend everyone to consider moving to Ruby 3 at the earliest.


codealchemy (approved):
👍

Kalimaha (approved):
LGTM 👍

PR #517, opened 2023-01-20
Add Ruby 3.2 to the CI matrix.

petergoldstein:
Runs green on my fork.

PR #524, opened 2023-02-05
Add Ruby 3.2 to CI

ecoologic:
Add support for Ruby 3.2 in CI.

Thank you to @petergoldstein , see https://github.com/zendesk/zendesk_api_client_rb/pull/517

Note that there is more work I need to do on the CHANGELOG, but I'm keeping this step separate.


Zendesk-NathanBellette (approved):
LGTM

Issue #513, opened 2023-01-12
Exceptions due to unexpected Faraday::Multipart::CompositeReadIO instances


jeremyjaybaker:
Hello, we just recently updated from 1.26 to 1.38.0.rc1 due to our other dependencies having a hard requirement of Faraday >=2.0. This seemed to go fine for the most part, but we're now getting the following exception for around <10% of our tickets and has not been reproducible outside of production.
```
NoMethodError: undefined method `key?' for #<Faraday::Multipart::CompositeReadIO:0x000055c1a8da7248>
  from zendesk_api/middleware/request/upload.rb:25:in `set_file'
  from zendesk_api/middleware/request/upload.rb:12:in `call'
  from zendesk_api/middleware/request/etag_cache.rb:22:in `call'
  from faraday/middleware.rb:17:in `call'
  from faraday/middleware.rb:17:in `call'
  from faraday/middleware.rb:17:in `call'
  from faraday/middleware.rb:17:in `call'
  from faraday/middleware.rb:17:in `call'
  from zendesk_api/middleware/response/parse_iso_dates.rb:11:in `call'
  from zendesk_api/middleware/response/logger.rb:22:in `call'
  from zendesk_api/middleware/response/callback.rb:14:in `call'
  from faraday/middleware.rb:17:in `call'
  from zendesk_api/middleware/response/raise_error.rb:8:in `call'
  from faraday/multipart/middleware.rb:28:in `call'
  from faraday/rack_builder.rb:153:in `build_response'
  from faraday/connection.rb:445:in `run_request'
  from faraday/connection.rb:281:in `post'
  from zendesk_api/actions.rb:32:in `save!'
  from zendesk_api/actions.rb:153:in `block in create!'
  from zendesk_api/actions.rb:152:in `tap'
  from zendesk_api/actions.rb:152:in `create!'
  from zendesk_api/collection.rb:64:in `block (2 levels) in <class:Collection>'
  from app/services/integrations/zendesk/zendesk_service.rb:74:in `create_ticket'
```

Our `ZendeskService#create_ticket` mentioned in the call stack calls the Zendesk API client directly like so:
```
client.tickets.create!(
  external_id: "12345foo",
  requester: #<ZendeskAPI::User>,
  organization: #<ZendeskAPI::Organization>,
  subject: "ACTION REQUIRED: Unable to process payment for Order #XXXXXX",
[...]

PR #526, opened 2023-03-03
[RED-1548] Fix issue with Faraday::Multipart

fbvilela:
cc @zendesk/redback 

### Description

This PR only removes the Faraday multipart middleware that was already being introduced further down the line in `client.rb`. It also adds a test in regards to the fix. 

## Links
[Issue](https://github.com/zendesk/zendesk_api_client_rb/issues/513)

Replicated issue:
replicated

Same test after the fix:
fixed

### DIFF
--- a/lib/zendesk_api/client.rb
+++ b/lib/zendesk_api/client.rb
@@ -143,8 +143,6 @@ module ZendeskAPI
     # Retry middleware if retry is true
     def build_connection
       Faraday.new(config.options) do |builder|
-        builder.request :multipart
-
         # response
         builder.use ZendeskAPI::Middleware::Response::RaiseError
         builder.use ZendeskAPI::Middleware::Response::Callback, self
--- a/spec/live/ticket_spec.rb
+++ b/spec/live/ticket_spec.rb
@@ -27,6 +27,41 @@ RSpec.describe ZendeskAPI::Ticket do
   it_should_be_readable agent, :ccd_tickets, create: true
   it_should_be_readable organization, :tickets
 
+  describe "#create" do
+    context "when passing large objects as parameters" do
+      let(:requester) { client.users.search(query: 'role:end-user').detect(&:photo) }
+      let(:organization) { client.organizations.sample }
+      let(:ticket_parameters) do
+        {
+          subject: 'live spec subject',
+          description: 'live spec description',
+          requester: requester,
+          organization: organization
+        } # We should always use requester/organiztion _id for existing records. This test should not be used as a guideline on how to use the sdk.
+      end
+
+      before do
+        VCR.use_cassette("ticket_create_with_large_objects") do
+          @ticket = ZendeskAPI::Ticket.create(client, ticket_parameters)
+        end
+      end
+
+      it 'is creatable' do
+        expect(requester).to_not be_nil
+
+        expect(@ticket.id).to_not be_nil
+        expect(@ticket.description).to eq(ticket_parameters[:description])
+      end
+    end
+
+    after do
+      return unless @ticket
+      VCR.use_cassette("ticket_destroy_with_large_objects") do
+        @ticket.destroy!
+      end
+    end
+  end
+
   describe "#attributes_for_save" do
     let :ticket do
       described_class.new(instance_double(ZendeskAPI::Client), status: :new)


ecoologic (commented):
Very nice work.

I think the tests are missing cleanup, ie deleting the created data. And the user selection could benefit from using search instead of getting them all.

Also, please link the issue in this PR, and when you're ready, update and close the issue.

spec/live/ticket_spec.rb
```diff
@@ -27,6 +27,39 @@ def valid_attributes
   it_should_be_readable agent, :ccd_tickets, create: true
   it_should_be_readable organization, :tickets
 
+  describe "#create" do
+    context "when passing large objects as parameters" do
+      let(:requester) { client.users.search(query: 'role:end-user').reject { |user| user.photo.nil? }.first }
```
ecoologic, line 32:
```suggestion
      let(:requester) { client.users.search(query: 'role:end-user').detect(&:photo) }
```

spec/live/ticket_spec.rb
```diff
@@ -27,6 +27,39 @@ def valid_attributes
...
+        expect(@ticket).to_not be_nil
+        expect(@ticket.send(:id)).to_not be_nil
+      end
+    end
+
+    after do
+      VCR.use_cassette("ticket_destroy_with_objects") do
+        @ticket.destroy!
+      end
+    end
```
ecoologic, line 60: If the ticket is not created we would have a confusing double error, which we can avoid with this:

```suggestion
    after do
      return unless @ticket
      VCR.use_cassette("ticket_destroy_with_objects") do
        @ticket.destroy!
      end
    end
```

spec/live/ticket_spec.rb
```diff
@@ -27,6 +27,39 @@ def valid_attributes
...
+        VCR.use_cassette("ticket_create_with_objects") do
+          @ticket = ZendeskAPI::Ticket.create(client, ticket_parameters)
+        end
+      end
+
+      it 'should be creatable' do
+        expect(requester).to_not be_nil
+        expect(@ticket).to_not be_nil
+        expect(@ticket.send(:id)).to_not be_nil
+      end
```
ecoologic, line 53: Using "should" in descriptions is not good practise, see https://www.betterspecs.org/#should 

Why are we using `send` here? That's not ideal for two reasons:

* `send` is for calling private methods
* even using `public_send` would be kinda confusing where `@ticket.id` would suffice

Is there any reason to use an instance var instead of a let here? Maybe something to do with `before and after`? See https://www.betterspecs.org/#let

Testing `not_nil` is a very weak test, can we add an expectation on the parameters being there? Also, when possible, it would be clearer to avoid negative statements, ref: https://www.obsidiansecurity.com/blog/why-to-stop-writing-negative-code/


```suggestion
      it 'creates the ticket' do
        expect(requester).to_not be_nil
        
        expect(ticket.id).to be_present
        expect(ticket.requested.reload!.photo.size).to be 123
      end
```

lib/zendesk_api/client.rb
```diff
@@ -143,8 +143,6 @@ def self.check_deprecated_namespace_usage(attributes, name)
     # Retry middleware if retry is true
     def build_connection
       Faraday.new(config.options) do |builder|
-        builder.request :multipart
```
tagpro, line 146: Why do we remove "multipart"?

tagpro (commented):
Erik's comments makes sense. I am not sure if I am the right person to contribute to this PR. Do have one question though in the comments.

lib/zendesk_api/client.rb
fbvilela: it's already present down the line, before the Upload middleware. this was a duplicate.

spec/live/ticket_spec.rb
fbvilela: thanks @ecoologic, I have pushed some changes. 
I couldn't use `be_present` in this case. it's added by Rails, hence the `not_nil`.
and yes the ticket needs to be an instance var because it gets created in the before block within that VCR code.

ecoologic (approved):
Very nice work 💯 

The GH issue can be closed.

PR #533, opened 2023-04-13
Re-run failed live specs

ecoologic:
Proof this is working: https://github.com/zendesk/zendesk_api_client_rb/actions/runs/4695301392/jobs/8324339290?pr=533

See how there's one failure, then the single spec re-runs green.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,6 +52,6 @@ jobs:
       - name: spec:live
         run: |
           bundle install
-          bundle exec rake clean_live set_ci_credentials spec:live
-          bundle exec rake clean_live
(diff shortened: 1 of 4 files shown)


fbvilela (approved):
very cool

.github/workflows/main.yml
```diff
@@ -52,4 +52,6 @@ jobs:
       - name: spec:live
         run: |
           bundle install
-          bundle exec rake clean_live set_ci_credentials spec:live
+          bundle exec rake clean_live set_ci_credentials spec:live ||
```
fbvilela, line 55: that's cool... I didn't know you could do that `||` here

PR #536, opened 2023-06-08
RED-1497 Ruby Client CBP by default

fbvilela:
Tries to make CBP requests when `page` is not specified. If the request raises an error it retries without the cbp specifics

### DIFF
--- a/lib/zendesk_api/collection.rb
+++ b/lib/zendesk_api/collection.rb
@@ -188,9 +188,8 @@ module ZendeskAPI
       elsif association && association.options.parent && association.options.parent.new_record?
         return (@resources = [])
       end
-      path_query_link = (@query || path)
 
-      get_resources(path_query_link)
+      get_resources(@query || path)
     end
 
     def fetch(*args)
@@ -250,6 +249,7 @@ module ZendeskAPI
         clear_cache
         @options["page"] = @options["page"].to_i + 1
       elsif (@query = @next_page)
+        # Send _only_ url param "?after=token" to get the next page
         @options.page&.delete("before")
         fetch(true)
       else
@@ -267,6 +267,7 @@ module ZendeskAPI
         clear_cache
         @options["page"] -= 1
       elsif (@query = @prev_page)
+        # Send _only_ url param "?before=token" to get the prev page
         @options.page&.delete("after")
         fetch(true)
       else
@@ -405,13 +406,7 @@ module ZendeskAPI
 
       while (bang ? fetch! : fetch)
         each do |resource|
-          arguments = [resource, @options["page"] || 1]
-
-          if block.arity >= 0
-            arguments = arguments.take(block.arity)
-          end
-
-          block.call(*arguments)
+          block.call(resource, @options["page"] || 1)
         end
 
         last_page? ? break : self.next
--- a/lib/zendesk_api/collection.rb
+++ b/lib/zendesk_api/collection.rb
@@ -344,7 +344,7 @@ module ZendeskAPI
     end
 
     def more_results?(response)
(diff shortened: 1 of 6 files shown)


lib/zendesk_api/collection.rb
```diff
@@ -272,6 +288,7 @@ def prev
         clear_cache
         @options["page"] -= 1
       elsif (@query = @prev_page)
+        @options.page&.delete("after")
```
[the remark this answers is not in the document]
fbvilela: the response will have both cursors under `links` and they will be both added to the `next|prev` request parameters if we don't delete the one that does not apply.

lib/zendesk_api/collection.rb
```diff
@@ -188,10 +190,24 @@ def fetch!(reload = false)
...
+      end
+
+      begin
+        @response = get_response(path_query_link)
+      rescue ZendeskAPI::Error::NetworkError => e
+        raise e if intentional_obp_request?
+        @options.per_page = @options_per_page_was
+        @options.page = nil
+        @response = get_response(path_query_link)
+      end
```
ecoologic, line 207: we're adding more than 10LOC in a method that's already long, we should extract this

lib/zendesk_api/collection.rb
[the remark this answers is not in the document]
ecoologic: I feel like a comment would help

lib/zendesk_api/collection.rb
```diff
@@ -351,17 +365,36 @@ def get_next_page_data(original_response_body)
 
     private
 
+    def cbp_response?(body)
+      body["meta"] && body["links"]
```
ecoologic, line 369: `!!`

spec/core/collection_spec.rb
```diff
@@ -5,6 +5,58 @@
     ZendeskAPI::Collection.new(client, ZendeskAPI::TestResource)
   end
 
+  # TODO!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+  # TODO: pager, xxx, url
```
ecoologic, line 9: will clean up, and I also need to write at least one test for `Helpers#present?`

lib/zendesk_api/collection.rb
```diff
@@ -351,17 +344,64 @@ def get_next_page_data(original_response_body)
...
+        @options.page = { size: (@options_per_page_was || DEFAULT_PAGE_SIZE) }
+      end
+
+      begin
+        @response = get_response(path_query_link)
+      rescue ZendeskAPI::Error::NetworkError => e
+        raise e if intentional_obp_request?
+        @options.per_page = @options_per_page_was
+        @options.page = nil
+        @response = get_response(path_query_link)
```
ecoologic, line 373: maybe better add a comment here.

lib/zendesk_api/collection.rb
```diff
@@ -351,17 +344,64 @@ def get_next_page_data(original_response_body)
...
+      end
+
+      if path_query_link == "search/export"
```
ecoologic, line 376: comment why this was left there.

lib/zendesk_api/collection.rb
```diff
@@ -370,13 +415,7 @@ def _all(start_page = @options["page"], bang = false, &block)
...
       while (bang ? fetch! : fetch)
         each do |resource|
-          arguments = [resource, @options["page"] || 1]
-
-          if block.arity >= 0
-            arguments = arguments.take(block.arity)
-          end
-
-          block.call(*arguments)
+          block.call(resource, @options["page"] || 1)
```
ecoologic, line 418: I can't think of any reason why these two versions should behave differently -> simplifying.

JinHuangMelb (approved):
LGTM

PR #543, opened 2023-07-13
[RED-1820] Do not attempt CBP requests on specific endpoints

fbvilela:
### Description
This PR is an attempt to reduce the number of requests that could potentially return 400 responses. 
`show_many` is one endpoint which we should not attempt any kind of pagination.


ecoologic (approved):
✅  I would have a few comments on readability, but I'm happy to hold them until we implement the bigger list, so we can push this out quickly.

The only thing is that it would be better to be strict on the matching and avoid `include?`. I'll follow up in video.

PR #546, opened 2023-07-18
[RED-1821] Remove CBP attempt and specify endpoints that support CBP

fbvilela:
### Description

Removes the `try CBP and retry with OBP if the first request fails` approach
Implements CBP on specific endpoints for the `Resource` ( whitelist approach )
Refactoring to improve readability in `collection.rb`

### DIFF
--- a/lib/zendesk_api/client.rb
+++ b/lib/zendesk_api/client.rb
@@ -37,24 +37,24 @@ module ZendeskAPI
     def method_missing(method, *args, &block)
       method = method.to_s
       options = args.last.is_a?(Hash) ? args.pop : {}
-
       unless config.use_resource_cache
-        raise "Resource for #{method} does not exist" unless method_as_class(method)
-        return ZendeskAPI::Collection.new(self, method_as_class(method), options)
+        resource_class = resource_class_for(method)
+        raise "Resource for #{method} does not exist" unless resource_class
+        return ZendeskAPI::Collection.new(self, resource_class, options)
       end
 
       @resource_cache[method] ||= { :class => nil, :cache => ZendeskAPI::LRUCache.new }
       if !options.delete(:reload) && (cached = @resource_cache[method][:cache].read(options.hash))
         cached
       else
-        @resource_cache[method][:class] ||= method_as_class(method)
+        @resource_cache[method][:class] ||= resource_class_for(method)
         raise "Resource for #{method} does not exist" unless @resource_cache[method][:class]
         @resource_cache[method][:cache].write(options.hash, ZendeskAPI::Collection.new(self, @resource_cache[method][:class], options))
       end
     end
 
     def respond_to?(method, *args)
-      ((cache = @resource_cache[method]) && cache[:class]) || !method_as_class(method).nil? || super
+      ((cache = @resource_cache[method]) && cache[:class]) || !resource_class_for(method).nil? || super
     end
 
     # Returns the current user (aka me)
@@ -184,7 +184,7 @@ module ZendeskAPI
 
     private
 
-    def method_as_class(method)
+    def resource_class_for(method)
       klass_as_string = ZendeskAPI::Helpers.modulize_string(Inflection.singular(method.to_s.gsub(/\W/, '')))
       ZendeskAPI::Association.class_from_namespace(klass_as_string)
     end
--- a/lib/zendesk_api/client.rb
+++ b/lib/zendesk_api/client.rb
@@ -54,7 +54,8 @@ module ZendeskAPI
     end
 
     def respond_to?(method, *args)
-      ((cache = @resource_cache[method]) && cache[:class]) || !resource_class_for(method).nil? || super
+      cache = @resource_cache[method]
+      !!(cache.to_h[:class] || resource_class_for(method) || super)
     end
 
     # Returns the current user (aka me)
@@ -185,8 +186,8 @@ module ZendeskAPI
     private
 
     def resource_class_for(method)
-      klass_as_string = ZendeskAPI::Helpers.modulize_string(Inflection.singular(method.to_s.gsub(/\W/, '')))
-      ZendeskAPI::Association.class_from_namespace(klass_as_string)
+      resource_name = ZendeskAPI::Helpers.modulize_string(Inflection.singular(method.to_s.gsub(/\W/, '')))
+      ZendeskAPI::Association.class_from_namespace(resource_name)
     end
 
     def check_url
--- a/lib/zendesk_api/collection.rb
+++ b/lib/zendesk_api/collection.rb
@@ -391,8 +391,14 @@ module ZendeskAPI
       @next_page, @prev_page = page_links(body)
 
       if cbp_response?(body)
-        @options["page"]["after"] = body["meta"]["after_cursor"]
-        @options["page"]["before"] = body["meta"]["before_cursor"]
+        @options.page = {} unless cbp_request?
+        # the line above means an intentional CBP request where page[size] is passed on the query
+        # this is to cater for CBP responses where we don't specify page[size] but the endpoint
+        # responds CBP by default. i.e  `client.trigger_categories.fetch`
+        @options.page.merge!(
+          after: body["meta"]["after_cursor"],
+          before: body["meta"]["before_cursor"]
+                             )
       elsif @next_page =~ /page=(\d+)/
         @options["page"] = $1.to_i - 1
       elsif @prev_page =~ /page=(\d+)/
--- a/lib/zendesk_api/collection.rb
+++ b/lib/zendesk_api/collection.rb
@@ -356,30 +356,27 @@ module ZendeskAPI
       Helpers.present?(@options["page"]) && !cbp_request?
     end
 
-    def should_try_cbp?(path_query_link)
-      not_supported_endpoints = %w[show_many]
-      not_supported_endpoints.none? { |endpoint| path_query_link.end_with?(endpoint) }
+    def supports_cursor_based_pagination?
+      @resource_class.const_defined?(:CBP_ACTIONS) && @resource_class.const_get(:CBP_ACTIONS).any? { |supported_path| path.end_with?(supported_path) }
+    end
+
+    def first_of_multiple_pages?
+      @next_page.nil?
+    end
+
+    def modify_request_for_cbp
+      @options_per_page_was = @options.delete("per_page")
+      # Default to CBP by using the page param as a map
+      @options.page = { size: (@options_per_page_was || DEFAULT_PAGE_SIZE) }
     end
 
     def get_resources(path_query_link)
       if intentional_obp_request?
         warn "Offset Based Pagination will be deprecated soon"
-      elsif @next_page.nil? && should_try_cbp?(path_query_link)
-        @options_per_page_was = @options.delete("per_page")
-        # Default to CBP by using the page param as a map
-        @options.page = { size: (@options_per_page_was || DEFAULT_PAGE_SIZE) }
+      elsif supports_cursor_based_pagination? && first_of_multiple_pages?
+        modify_request_for_cbp
       end
-
-      begin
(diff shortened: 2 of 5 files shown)


lib/zendesk_api/client.rb
```diff
@@ -37,10 +37,11 @@ class Client
...
       method = method.to_s
       options = args.last.is_a?(Hash) ? args.pop : {}
-
+      resource_class = method_as_class(method)
+      options = options.merge(resource_class.default_collection_options) if resource_class.respond_to?(:default_collection_options)
       unless config.use_resource_cache
-        raise "Resource for #{method} does not exist" unless method_as_class(method)
-        return ZendeskAPI::Collection.new(self, method_as_class(method), options)
+        raise "Resource for #{method} does not exist" unless resource_class
+        return ZendeskAPI::Collection.new(self, resource_class, options)
```
ecoologic, line 44: Method missing has a huge footprint and this requires some extra confidence, seems fine, but I might want to dig deeper.

ecoologic (commented):
Preliminary review of DRAFT PR:

lib/zendesk_api/resources.rb
```diff
@@ -385,6 +385,17 @@ def attribute_changes
       attributes.changes.merge("comment" => attributes["comment"])
     end
 
+    def self.show_many(client, options = {})
+      # this is no longer needed... I just wanted to explore different ways of doing things and
+      # how to override magic methods e.t.c
+      # we no longer need to say attempt_cbp: false - since false is now the default behaviour
+      Collection.new(client, ZendeskAPI::Ticket, options.merge(attempt_cbp: false , :path => "tickets/show_many", :verb => :get))
```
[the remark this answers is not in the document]
fbvilela: I will delete this code. I was just exploring this way of "blacklisting" 
but now it's on a whitelist mechanism so the show_many would be false by default

lib/zendesk_api/client.rb
```diff
@@ -37,10 +37,11 @@ class Client
     def method_missing(method, *args, &block)
       method = method.to_s
       options = args.last.is_a?(Hash) ? args.pop : {}
-
+      resource_class = method_as_class(method)
```
ecoologic, line 40: Great naming 💯

spec/live/ticket_spec.rb
```diff
@@ -62,6 +62,23 @@ def valid_attributes
     end
   end
 
+  describe "#show_many" do
+    let(:how_many_tickets) { 10 }
+    before(:all) do
```
ecoologic, line 67: `before all ` is a bit of a code smell that tends to create order dependent specs. there is only one spec, just `before` which defaults to `:each` should suffice.

lib/zendesk_api/collection.rb
```diff
@@ -357,14 +357,13 @@ def intentional_obp_request?
     end
 
     def should_try_cbp?(path_query_link)
-      not_supported_endpoints = %w[show_many]
-      not_supported_endpoints.none? { |endpoint| path_query_link.end_with?(endpoint) }
+      @next_page.nil? && @options.delete(:attempt_cbp)
```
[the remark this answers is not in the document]
ecoologic: It would be nice to explain why `@next_page.nil?`, in a smaller class this could be extracted to an intention revealing method, but maybe just a local var or a comment would work in this case.

lib/zendesk_api/collection.rb
```diff
@@ -357,14 +357,13 @@ def intentional_obp_request?
...
     end
 
     def get_resources(path_query_link)
       if intentional_obp_request?
         warn "Offset Based Pagination will be deprecated soon"
-      elsif @next_page.nil? && should_try_cbp?(path_query_link)
+      elsif should_try_cbp?(path_query_link)
```
ecoologic, line 366: If we remove the two calls, then there's no "try", we know which XBP to use

lib/zendesk_api/client.rb
ecoologic: makes me think that `method_as_class` could be renamed to `resource_class_for(method)`

lib/zendesk_api/collection.rb
```diff
@@ -541,9 +542,13 @@ def array_method(name, *args, &block)
       to_a.public_send(name, *args, &block)
     end
 
+    # If you call client.tickets.foo - and foo is not an attribute nor an association, it ends up here, as a new collection
     def next_collection(name, *args, &block)
       opts = args.last.is_a?(Hash) ? args.last : {}
-      opts.merge!(:collection_path => @collection_path.dup.push(name))
+      opts.merge!(collection_path: @collection_path.dup.push(name), page: nil)
```
ecoologic, line 548: Minor:

```suggestion
      opts.merge!(collection_path: [*@collection_path, name], page: nil)
```

It's more declarative (no `dup.push`).

lib/zendesk_api/collection.rb
```diff
@@ -395,8 +392,12 @@ def set_page_and_count(body)
 
       if cbp_response?(body)
         # We'll delete the one we don't need in #next or #prev
-        @options["page"]["after"] = body["meta"]["after_cursor"]
-        @options["page"]["before"] = body["meta"]["before_cursor"]
+        if @options.page.is_a?(Hash)
```
ecoologic, line 395:
```suggestion
        if cbp_request?
```

This needs an explanation, and the comment above might live better in the next line now.

ecoologic (commented):
A few minor points for you to consider. ❤️ 

Good job Fabio!

lib/zendesk_api/collection.rb
fbvilela: HA! the code is slightly different now... but here's the reason why: 
I found out that two live specs started to fail because even though the code wasn't doing an explicit CBP request, the response was CBP, containing the `has_more` and before/after cursors. One of the cases is `api/v2/trigger_categories.json` . In the previous version of this gem, you could make the request with `client.trigger_categories.fetch` but you would never go past the first page.

fbvilela: I wrote a test for this case, it's in a separate commit.

lib/zendesk_api/collection.rb
```diff
@@ -356,30 +356,27 @@ def intentional_obp_request?
...
-      not_supported_endpoints.none? { |endpoint| path_query_link.end_with?(endpoint) }
+    def supports_cursor_based_pagination?
+      @resource_class.const_defined?(:CBP_ACTIONS) && @resource_class.const_get(:CBP_ACTIONS).any? { |supported_path| path.end_with?(supported_path) }
+    end
+
+    def first_of_multiple_pages?
+      @next_page.nil?
+    end
+
+    def modify_request_for_cbp
```
ecoologic, line 367: This is good.

Minor: as a private method, I wonder if `set_cbp_options` would be more explicit.

lib/zendesk_api/collection.rb
```diff
@@ -356,30 +356,27 @@ def intentional_obp_request?
       Helpers.present?(@options["page"]) && !cbp_request?
     end
 
-    def should_try_cbp?(path_query_link)
-      not_supported_endpoints = %w[show_many]
-      not_supported_endpoints.none? { |endpoint| path_query_link.end_with?(endpoint) }
+    def supports_cursor_based_pagination?
```
ecoologic, line 359: This is good.

Minor: Considering we're using cbp and obp freely, I wonder if `supports_cbp?` would be better.

lib/zendesk_api/collection.rb
```diff
@@ -356,30 +356,27 @@ def intentional_obp_request?
...
+      @resource_class.const_defined?(:CBP_ACTIONS) && @resource_class.const_get(:CBP_ACTIONS).any? { |supported_path| path.end_with?(supported_path) }
+    end
+
+    def first_of_multiple_pages?
```
ecoologic, line 363: Not bad.

* Is this for the first page only?
* Do we know there are multiple pages?

Can you add a method comment to explain why `@next_page.nil?` defines the first of multiple pages?

lib/zendesk_api/collection.rb
```diff
@@ -394,9 +391,14 @@ def set_page_and_count(body)
...
-        @options["page"]["after"] = body["meta"]["after_cursor"]
-        @options["page"]["before"] = body["meta"]["before_cursor"]
+        @options.page = {} unless cbp_request?
+        # the line above means an intentional CBP request where page[size] is passed on the query
+        # this is to cater for CBP responses where we don't specify page[size] but the endpoint
+        # responds CBP by default. i.e  `client.trigger_categories.fetch`
+        @options.page.merge!(
+          after: body["meta"]["after_cursor"],
+          before: body["meta"]["before_cursor"]
+                             )
```
ecoologic, line 401: Can we extract this?

Could we have `before` before after?

spec/live/ticket_spec.rb
```diff
@@ -62,6 +62,23 @@ def valid_attributes
     end
   end
 
+  describe "#show_many" do
+    let(:how_many_tickets) { 10 }
+    before do
+      VCR.use_cassette("get_tickets_show_many") do
+        @ticket_ids = client.tickets.per_page(how_many_tickets).fetch.map(&:id)
+      end
```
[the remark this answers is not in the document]
fbvilela: they are both real tickets... Instances of ZendeskAPI::Ticket so I thought it wouldn't make much sense to rename one of them to `actual_tickets`

lib/zendesk_api/collection.rb
```diff
@@ -541,9 +538,13 @@ def array_method(name, *args, &block)
...
     end
 
+    # If you call client.tickets.foo - and foo is not an attribute nor an association, it ends up here, as a new collection
     def next_collection(name, *args, &block)
       opts = args.last.is_a?(Hash) ? args.last : {}
-      opts.merge!(:collection_path => @collection_path.dup.push(name))
+      opts.merge!(collection_path: [*@collection_path, name], page: nil)
+      # why page: nil ?
+      # when you do client.tickets.fetch followed by client.tickets.foos => the request to /tickets/foos will
+      # have the options page set to whatever the last options were for the tickets collection
```
ecoologic, line 547: This is a great explanation. 🥇

spec/live/ticket_spec.rb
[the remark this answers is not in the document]
ecoologic: OK thanks

lib/zendesk_api/collection.rb
```diff
@@ -348,6 +348,12 @@ def cbp_response?(body)
       !!(body["meta"] && body["links"])
     end
 
+    def set_cbp_options
+      @options_per_page_was = @options.delete("per_page")
+      # Default to CBP by using the page param as a map
+      @options.page = { size: (@options_per_page_was || DEFAULT_PAGE_SIZE) }
+    end
+
     def cbp_request?
```
ecoologic, line 357: Suggestion:

```suggestion
    # CBP requests look like: `/resources?page[size]=100`
    # OBP requests look like: `/resources?page=2`
    def cbp_request?
```

lib/zendesk_api/collection.rb
```diff
@@ -356,30 +362,23 @@ def intentional_obp_request?
       Helpers.present?(@options["page"]) && !cbp_request?
     end
 
-    def should_try_cbp?(path_query_link)
-      not_supported_endpoints = %w[show_many]
-      not_supported_endpoints.none? { |endpoint| path_query_link.end_with?(endpoint) }
+    def supports_cbp?
+      @resource_class.const_defined?(:CBP_ACTIONS) && @resource_class.const_get(:CBP_ACTIONS).any? { |supported_path| path.end_with?(supported_path) }
```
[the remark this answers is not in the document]
fbvilela: I'm suggesting to leave this as is until the next PR which will implement more endpoints and regex

ecoologic (approved):
Great work Fabio ✅

PR #548, opened 2023-07-26
[RED-1821] Cater for endpoints with CBP support

fbvilela:
### Description

Adding endpoints that support CBP.


spec/live/cbp_support.rb
```diff
@@ -0,0 +1,37 @@
...
+describe 'Endpoints that support CBP' do
+  describe ZendeskAPI::Group do
+    describe '/groups' do
+      expect_cbp_response_for(client.groups)
+      expect_cbp_response_for(client.groups.assignable)
+    end
+  end
+
+  describe ZendeskAPI::GroupMembership do
+    describe '/groups/{id}/memberships' do
```
fbvilela, line 12: sigh, there's some code repetition here... since the test makes two requests instead of one.. .not sure how to get around this... still thinking about it

spec/live/cbp_support.rb
```diff
@@ -0,0 +1,37 @@
...
+      before do
+        VCR.use_cassette("cbp_group_memberships_all_groups") do
+          @groups = client.groups.fetch
```
ecoologic, line 15: Could we use `let` instead of instance vars?

https://www.betterspecs.org/#let

spec/core/cbp_helper.rb
```diff
@@ -0,0 +1,21 @@
+def expect_cbp_response_for(collection)
+  context collection.path.to_s, :vcr do
+    before do
+      @resource_klass = collection.instance_variable_get(:@resource_class)
+      VCR.use_cassette("cbp_#{@resource_class}_#{collection.path}") do
+        @result = collection.fetch
+        @response_body = collection.response.body
+      end
+    end
+    it 'expects an array with the correct element types' do
```
ecoologic, line 10: Maybe?

```suggestion
    it "returns a list of #{@resource_klass}" do
```

I think in this context it should be available, you're inside a method....

However, I'm not a fan of extracting whole tests in methods. But I can't verbalize that, so maybe it's a subjective thing. I'm ok with a bit of duplication in the code, or using a list of test params like we do in Go.

The context could also use some more wording.

Lastly, I see the context, but I don't see what are we describing?

spec/core/cbp_helper.rb
```diff
@@ -0,0 +1,21 @@
+def expect_cbp_response_for(collection)
+  context collection.path.to_s, :vcr do
+    before do
+      @resource_klass = collection.instance_variable_get(:@resource_class)
```
ecoologic, line 4: Any way around poking inside the private variables of classes?

ecoologic (approved):
✅  This is great work, see my latest comments and if there's anything you'd like to implement. When you're happy with it, I'm happy too.

spec/live/cbp_support.rb
fbvilela: Done it ✅

spec/core/cbp_helper.rb
fbvilela: Done it ✅ 
also using the described_class var provided by rspec, much nicer

spec/core/cbp_helper.rb
fbvilela: yes, this was taken care of

lib/zendesk_api/collection.rb
```diff
@@ -365,7 +365,8 @@ def intentional_obp_request?
     end
 
     def supports_cbp?
-      @resource_class.const_defined?(:CBP_ACTIONS) && @resource_class.const_get(:CBP_ACTIONS).any? { |supported_path| path.end_with?(supported_path) }
+      @resource_class.const_defined?(:CBP_ACTIONS) &&
+        @resource_class.const_get(:CBP_ACTIONS).any? { |supported_path_regex| path.match?(supported_path_regex) }
```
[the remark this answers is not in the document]
fbvilela: Done ✅ 
I created a module for the CBP support code. was that what you were thinking?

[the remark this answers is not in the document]
ecoologic: no... my idea was to create a method in resource, returning an empty array, and override the method in the specific resources.

I'm ok with what we have but if you're curious we can chat about it 👍

lib/zendesk_api/cbp_support.rb
```diff
@@ -0,0 +1,17 @@
...
+    end
+
+    # Returns the paths that support CBP
+    # @return [Array] of regular expressions
+    # To add CBP support to a resource, add a constant called CBP_ACTIONS. See examples in resources.rb
+    module ClassMethods
+      def cbp_path_regexes
+        const_defined?(:CBP_ACTIONS) ? const_get(:CBP_ACTIONS) : []
+      end
+    end
```
ecoologic, line 15: I think this overcomplicates things, too much indirection.

lib/zendesk_api/resources.rb
```diff
@@ -378,7 +382,7 @@ class Ticket < Resource
     extend UpdateMany
     extend DestroyMany
 
-    CBP_ACTIONS = %w[tickets].freeze
+    CBP_ACTIONS = [/tickets$/].freeze
```
ecoologic, line 385:
```suggestion
    def self.cbp_actions
      [/tickets$/].freeze
    end
```

lib/zendesk_api/collection.rb
```diff
@@ -568,16 +479,5 @@ def assert_results(results, body)
...
-    def set_cbp_response_options(body)
-      @options.page = {} unless cbp_request?
-      # the line above means an intentional CBP request where page[size] is passed on the query
-      # this is to cater for CBP responses where we don't specify page[size] but the endpoint
-      # responds CBP by default. i.e  `client.trigger_categories.fetch`
-      @options.page.merge!(
-        before: body["meta"]["before_cursor"],
-        after: body["meta"]["after_cursor"]
-                           )
-    end
```
ecoologic, line 581: It's always great to see files get smaller 😄

lib/zendesk_api/pagination.rb
```diff
@@ -0,0 +1,107 @@
...
+    DEFAULT_PAGE_SIZE = 100
+    def more_results?(response)
+      Helpers.present?(response["meta"]) && response["meta"]["has_more"]
+    end
+    alias has_more_results? more_results? # For backward compatibility with 1.33.0 and 1.34.0
+
+    def each_page!(*args, &block)
+      warn "ZendeskAPI::Collection#each_page! is deprecated, please use ZendeskAPI::Collection#all!"
+      all!(*args, &block)
+    end
```
ecoologic, line 13: this (and below) is not pagination logic?

lib/zendesk_api/pagination.rb
```diff
@@ -0,0 +1,107 @@
+module ZendeskAPI
+  # Contains all methods related to pagination in an attempt to slim down collection.rb
```
ecoologic, line 2: Minor: If you wanted to be super-clear about the fact it's only for collection, you could even namespace it by `module Collection` here. That way, it wouldn't pollute the namespace in `ZendeskAPI` when it should only be used in `Collection`.

lib/zendesk_api/pagination.rb
```diff
@@ -0,0 +1,107 @@
...
+      # the line above means an intentional CBP request where page[size] is passed on the query
+      # this is to cater for CBP responses where we don't specify page[size] but the endpoint
+      # responds CBP by default. i.e  `client.trigger_categories.fetch`
+      @options.page.merge!(
+        before: body["meta"]["before_cursor"],
+        after: body["meta"]["after_cursor"]
+                           )
+    end
+  end
+end
```
ecoologic, line 107: Nice! Scrolling through the file you can kinda understand how pagination is handled. 👍

lib/zendesk_api/pagination.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    def first_page?
+      !@prev_page
+    end
+
+    def last_page?
+      !@next_page || @next_page == @query
+    end
+
```
ecoologic, line 43: Moving public methods make is a bit harder to follow in case they're called from outside of Collection, but in this case I'm fine with it. 👍

lib/zendesk_api/resources.rb
```diff
@@ -650,6 +664,18 @@ class GroupMembership < Resource
...
     has User
     has Group
+
+    def self.cbp_path_regexes
+      [%r{groups/\d+/memberships$}]
+    end
+  end
+
+  class Group < Resource
+    has_many :memberships, :class => GroupMembership, :path => "memberships"
```
ecoologic, line 674: Very minor, but why not? 😄 

```suggestion
    has_many :memberships, class: GroupMembership, path: "memberships"
```

spec/live/cbp_support_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'core/spec_helper'
+require 'core/resources/cbp_spec_helper'
+
+describe 'Endpoints that support CBP' do
+  describe ZendeskAPI::Group do
+    describe '/groups' do
```
ecoologic, line 6: I think RSpec is about describing `Object#methods`. This makes the test suite into documentation. IE: `describe "#fetch"`

https://www.betterspecs.org/#describe

PR #551, opened 2023-08-03
Add TargetFailure

fulcanelly:
https://developer.zendesk.com/api-reference/ticketing/targets/target_failures/#show-target-failure

### DIFF
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -500,6 +500,14 @@ module ZendeskAPI
     put :recover
   end
 
+  class TargetFailure < ReadResource
+    def target
+      @client.targets.to_a.find do |target|
+        target.title == self.target_name
+      end
+    end
+  end
+
   class DeletedTicket < ReadResource
     include Destroy
     extend DestroyMany


lib/zendesk_api/resources.rb
```diff
@@ -500,6 +500,14 @@ class SuspendedTicket < ReadResource
     put :recover
   end
 
+  class TargetFailure < ReadResource
+    def target
+      @client.targets.to_a.find do |target|
```
ecoologic, line 505: If you use `.all` in stead of `to_a` you should be able to reduce the number of calls in case of pagination.

lib/zendesk_api/resources.rb
```diff
@@ -500,6 +500,14 @@ class SuspendedTicket < ReadResource
...
+        target.title == self.target_name
```
ecoologic, line 506: Given the match is done on title, I would suggest naming the method `target_by_title`. (?? not too sure).

ecoologic (commented):
Hi @fulcanelly. Thank you for your contribution.

A couple of comments, but mainly, we can't approve any work that doesn't **come with tests**.

Could you write some unit tests for this?

Thanks again

PR #552, opened 2023-08-21
Red 1856 equality cleanup

ecoologic:


ecoologic:
There's a better way.

PR #555, opened 2023-08-21
Refactor Side Load of `ZendeskAPI::Association` Into A Separate Object

khezam:
It seems like the side load behaviors of the `ZendeskAPI::Association` class could be encapsulated in a separate object for extendability so I thought it might be a good opportunity to refactor.

Let me know if it's not needed and I'm happy to close this PR.

### DIFF
--- a/lib/zendesk_api/association.rb
+++ b/lib/zendesk_api/association.rb
@@ -73,20 +73,78 @@ module ZendeskAPI
       namespace.join("/")
     end
 
-    # Tries to place side loads onto given resources.
-    def side_load(resources, side_loads)
-      key = "#{options.name}_id"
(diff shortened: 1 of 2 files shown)


ecoologic:
Hi @khezam  - thank you for another contribution! (your previous one is coming, I just need another internal review).

We don't have much time at present to do a deep review on it, but eventually we'll find the time.

First question: Is this code already covered by existing tests or would it benefit from extra testing?

Thanks again ❤️

khezam:
Hi @ecoologic - my pleasure and I'm happy I could help :) 

> We don't have much time at present to do a deep review on it, but eventually we'll find the time.

No worries at all and I totally understand that.

> First question: Is this code already covered by existing tests or would it benefit from extra testing?

Yes, this code is already covered by existing tests  :)

PR #556, opened 2023-08-22
[RED-1886] Add more CBP endpoints 

fbvilela:
### Description

Adds a few more endpoints to the list of supported CBP

### Risk
* low


khezam (approved):
This looks good to me 👍 💯  A small optional suggestion: It might be more consistent to keep class methods in `class << self` or `self.method_name`.

ecoologic (approved):
✅  - I'd love it if we had a link to the api doc for every regex. But no biggie.

Good job on the shared examples 👍

PR #566, opened 2024-05-09
Add Gemfile.lock for different Ruby versions

JinHuangMelb:
Add Gemfile for different Ruby versions.
Add Gemfile.lock file for Snyk scan.


ecoologic (approved):
very nice! ✅

PR #569, opened 2024-05-20
Fix live specs

ecoologic:


Gemfile.2.7.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    zendesk_api (3.0.5)
+    zendesk_api (3.0.6)
```
ecoologic, line 4: It appears to me, the tests were running the old version!!

This is true for all lock files. I think the automation is not working. `!!!`

PR #568, opened 2024-05-20
Bump v3.0.6 - Add Gemfile and Gemfile.lock for different Ruby versions

JinHuangMelb:
Release `v3.0.6`

[Diff](https://github.com/zendesk/zendesk_api_client_rb/compare/v3.0.5...v3.0.6)

Ref. RED-2160

PR #570, opened 2024-05-20
Revert "Bump v3.0.6 - Add Gemfile and Gemfile.lock for different Ruby versions"

JinHuangMelb:
Reverts zendesk/zendesk_api_client_rb#568

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 # CHANGELOG
 
-## v3.0.6
-
-Adding multiple Gemfile and Gemfile.lock files for differen Ruby versions so that Snyk can scan the repo for vulnerabilities.
-
 ## v3.0.5
 
 Adding CBP support to more endpoints. please see `cbp_support_spec.rb` for the complete list.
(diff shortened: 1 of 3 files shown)


ecoologic (commented):
I don't think we should revert versions that have already been made available to the public. We want all the `3.0.6` versions to be exactly the same code base.

It might be argued that this is only CI changes, so it was redundant to release a version in the first place, but what I would do at this point is just leave the main branch as is (closing this PR), and when we ship some value to the customers we can release a higher version.

Lastly, if we really wanted to undo a version release, we would have to delete the git tag and remove the version from rubygems, something I would do only if no-one ever downloaded the new version. -- Now I checked and noticed that we haven't even released 3.0.5 anyway [in rubygems](https://rubygems.org/search?query=zendesk_api), so seems like the release process wasn't even followed to completion. 🤔  I don't think this changes my opinion about this PR. I suggest closing and moving on.

JinHuangAtZen:
> I don't think we should revert versions that have already been made available to the public. We want all the `3.0.6` versions to be exactly the same code base.
> 
> It might be argued that this is only CI changes, so it was redundant to release a version in the first place, but what I would do at this point is just leave the main branch as is (closing this PR), and when we ship some value to the customers we can release a higher version.
> 
> Lastly, if we really wanted to undo a version release, we would have to delete the git tag and remove the version from rubygems, something I would do only if no-one ever downloaded the new version. -- Now I checked and noticed that we haven't even released 3.0.5 anyway [in rubygems](https://rubygems.org/search?query=zendesk_api), so seems like the release process wasn't even followed to completion. 🤔 I don't think this changes my opinion about this PR. I suggest closing and moving on.

Yup, I didn't go ahead to release 3.0.5 because the issue we found, will close this PR for now.

Issue #581, opened 2024-08-14
Offset Based Pagination will be deprecated soon


navidemad:
# Context
rails: 7.1.3.4
zendesk_api: 3.1.0
ruby: 3.3.4

# Goal
Fetching requested tickets from our customer

# Problem
We are having the following warning when we followed the CBP example from:
https://github.com/zendesk/zendesk_api_client_rb?tab=readme-ov-file#pagination

> Offset Based Pagination will be deprecated soon

# Debug
Our customer has 10 tickets.
I used per_page(2) to make sure the pagination works fine, we will put 100 once we fixed this issue.
I put some log on https://github.com/search?q=repo%3Azendesk%2Fzendesk_api_client_rb%20intentional_obp_request%3F&type=code

```
@options: #<ZendeskAPI::SilentMash page=nil query="email:flabro@hotmail.com">
@options: #<ZendeskAPI::SilentMash per_page=2>
@options: #<ZendeskAPI::SilentMash page=2 per_page=2>
Offset Based Pagination will be deprecated soon
@options: #<ZendeskAPI::SilentMash page=3 per_page=2>
Offset Based Pagination will be deprecated soon
@options: #<ZendeskAPI::SilentMash page=4 per_page=2>
Offset Based Pagination will be deprecated soon
@options: #<ZendeskAPI::SilentMash page=5 per_page=2>
Offset Based Pagination will be deprecated soon
```

# Code:

```ruby
# frozen_string_literal: true

class Zendesk::AllRequestedTicketsService
  attr_reader :errors

  def initialize(user_id:)
    @user_id = user_id
    @errors = []
  end

  def success?
    @errors.blank?
  end

  def failure?
    @errors.present?
  end

  # https://github.com/zendesk/zendesk_api_client_rb?tab=readme-ov-file#pagination
  def call
    user = User.find_by(id: @user_id)
    return [] if user.blank?

    zendesk_client = ZendeskAPI::Client.new do |config|
      config.url = ENV.fetch("ZENDESK_URL")
      config.username = ENV.fetch("ZENDESK_USERNAME")
      config.token = ENV.fetch("ZENDESK_TOKEN")
      config.retry = true
    end

    zendesk_user_client = zendesk_client.users.search(query: "email:#{user.email}").first
    return [] if !zendesk_user_client || !zendesk_user_client.respond_to?(:id) || zendesk_user_client.id.blank?

    tickets = Set.new
[...]

PR #582, opened 2024-08-19
[RED-2366] Add CBP support to requested_tickets api/v2/users/:id/tickets/requested

fbvilela:
### Description

Adding CBP support to this api endpoint 

### References 

 #>https://github.com/zendesk/zendesk_api_client_rb/issues/581

### DIFF
--- a/spec/macros/resource_macros.rb
+++ b/spec/macros/resource_macros.rb
@@ -110,7 +110,7 @@ module ResourceMacros
           if options[:find]
             expect(obj.send(options[:find].first)).to eq(options[:find].last)
           else
-            expect(obj.nil? || !obj.active?).to eq(true)
+            options[:marked_for_deletion] ? (expect(obj.active?).to be_falsey) : (expect(obj).to be_nil)
           end
         end
       end
--- a/spec/macros/resource_macros.rb
+++ b/spec/macros/resource_macros.rb
@@ -37,7 +37,7 @@ module ResourceMacros
       end
 
       after(:all) do
-        return unless @creatable_object.id
+        return unless @creatable_object&.id
 
         VCR.use_cassette("#{described_class.to_s}_create_delete") do
           @creatable_object.destroy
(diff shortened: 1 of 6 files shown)


spec/macros/resource_macros.rb
```diff
@@ -41,6 +41,7 @@ def it_should_be_creatable(options = {})
 
         VCR.use_cassette("#{described_class.to_s}_create_delete") do
           @creatable_object.destroy
+          sleep(1)
```
ecoologic, line 44: A consideration, maybe not even for this PR:

if the object gets returned even after deletion, maybe it has an attribute `deleted: true` or something we can check and make the test pass?

spec/macros/resource_macros.rb
```diff
@@ -110,7 +114,7 @@ def it_should_be_deletable(options = {})
           if options[:find]
             expect(obj.send(options[:find].first)).to eq(options[:find].last)
           else
-            expect(obj).to be_nil
+            expect(obj.nil? || !obj.active?).to eq(true)
```
ecoologic, line 117: can you explain what's going on here?

caspersg: Can you (@fbvilela), test if the `random_string` fix works without this change?

fbvilela: Casper, it didn't work. 

This change is because `OrganizationField` is marked for deletion and returned with an `active: false ` . It eventually gets deleted for real. I will reach out to the team but as for our code change here, I might try to make this specific for the organization field class.

caspersg: > I will reach out to the team 

Yeah this sounds like a bug to me, as discussed in slack.

> make this specific for the organization field class

Yeah, if we keep this change, it should be specific to the case that is affected at least.

token-cjg (approved):

caspersg (approved):

PR #590, opened 2025-08-12
drop support for Ruby 2.7 and 3.0, update CI

razumau:
Drops support for EOL versions of Ruby (even 3.1 is actually EOL at this time, but we’ll leave it for now).

Checks in `Gemfile.lock` (I spent some time figuring out why some tests failed on Ruby 2.7 but not 3.4, and that was because different versions of faraday got installed for 2.7 and 3.4).

Updates rubocop and applies a bunch of cops.

In CI workflow: uses correct main branch, updates checkout and setup-ruby actions, runs spec-live on Ruby 3.4, adds JRuby 10, removes unnecessary `bundle install`.

To do in follow-up PRs: remove limits on rspec gems, simplify `Gemfile`.


codealchemy (approved):

.github/workflows/main.yml
```diff
@@ -5,34 +5,33 @@ on:
...
           - "3.1"
           - "3.2"
           - "3.3"
           - "3.4"
           - "jruby-9.4"
+          - "jruby-10.0"
     steps:
       - name: Checkout code
-        uses: zendesk/checkout@v2
+        uses: actions/checkout@v4
```
bquorning, line 27:
```suggestion
        uses: actions/checkout@v5
```

zendesk_api.gemspec
```diff
@@ -18,18 +18,19 @@ Gem::Specification.new do |s|
...
     'source_code_uri' => "https://github.com/zendesk/zendesk_api_client_rb/tree/v#{s.version}",
-    'wiki_uri' => 'https://github.com/zendesk/zendesk_api_client_rb/wiki'
+    'wiki_uri' => 'https://github.com/zendesk/zendesk_api_client_rb/wiki',
+    'rubygems_mfa_required' => 'true'
   }
 
   s.files = Dir.glob('{lib,util}/**/*') << 'LICENSE'
 
-  s.required_ruby_version     = ">= 2.7"
+  s.required_ruby_version     = ">= 3.1"
```
bquorning, line 27: Do we know (do we have a way of knowing) which version our users are using?

PR #593, opened 2025-10-08
add support for refresh tokens

zendesk-mradmacher:
# Add support for refreshing OAuth access token.

When you need some inspiration, consult how [the token refresh flow](https://github.com/restforce/restforce/blob/main/lib/restforce/middleware/authentication.rb#L24) is done by [Restforce](https://github.com/restforce/restforce), the public gem for Salesforce REST API.

I add a new `ZendeskAPI::TokenRefresher` service, to obtain new access and refresh tokens. It takes client configuration as a parameter. When tokens get refreshed, the client configuration is updated, so the next requests are issued using the newly obtained access token. Also a block is yielded with new access and refresh tokens, so they could be stored for further use.

The alternative solution is to use `ZendeskAPI::Middleware::Response::TokenRefresher`, to start the process of token refreshing each time when the API responds with 401. The middleware also updates the client configuration. It requires a callback to be configured. The called is called whenever new tokens are obtained to allow storing them for further use.
There is a configuration option to enable middleware usage.

### How to test?
In Admin Center, Apps and Integrations / APIs / OAuth Clients create a new client. Note client id and secret and redirect URL. The redirect URL can be any address.

- Login to Zendesk as an user.
- Assumming your Zendesk subdomain is "testing", open the following URL in browser:
```
https://testing.zendesk.com/oauth/authorizations/new?client_id=<your client id>&response_type=code&redirect_uri=<encoded redirect URL>&scope=read
```
The encoded URL looks like https%3A%2F%2Fexample.com.
- Accept the authorization and intercept the code issued to the redirect URL.
- Request for access and refresh tokens using the intercepted code:
```
curl https://testing.zendesk.com/oauth/tokens \
  -H "Content-Type: application/json" \
  -d '{"grant_type": "authorization_code", "code": "<here goes the code>", "client_id": "<your client id>", "client_secret": "<your client secret>", "redirect_uri": "<your redirect URL", "scope": "read", "expires_in": 300 }' \
  -X POST
```
`expires_in` is chosen to be 5 minutes (the lowest possible value), to make the access token expire quickly, for easier testing.
- Save tokens in a file named `tokens` in the following format:
```
<access token>,<refresh token>
```
- Use the following Ruby script to make requests to your Zendesk. Observe how tokens are refreshed upon expiring.
```
require 'zendesk_api'

# `tokens` is a one line file that stores comma separated tokens:
# <access token>,<refresh token>
access_token, refresh_token = File.open("tokens") do |file|
  file.read.strip.split(",")
end

client = ZendeskAPI::Client.new do |config|
  # Mandatory:
  config.url = "https://testing.zendesk.com/api/v2"

  config.client_id = "<your client id>"
  config.client_secret = "<your client secret>"
  config.access_token_expiration = 300 # expire quickly to observe token refreshing

  config.access_token = access_token
  config.refresh_token = refresh_token

  # In case of using the middleware for refreshing tokens, update the storage with new tokens
  # config.auto_refresh_tokens = true
  # config.refresh_tokens_callback = lambda do |access_token, refresh_token|
  #   File.open("tokens", "w") do |file|
  #     file.write("#{access_token},#{refresh_token}")
  #   end
  # end

  require 'logger'
  config.logger = Logger.new(STDOUT)
end

users = client.users.per_page(1)
begin
  # A request with an expired access token is made.
  users.fetch!
  # The request is rejected with 401 status code.
rescue ZendeskAPI::Error::Unauthorized
  # Refresh tokens and store them securely
  ZendeskAPI::TokenRefresher.new(client.config).refresh_token do |access_token, refresh_token|
    File.open("tokens", "w") do |file|
      file.write("#{access_token},#{refresh_token}")
    end
  end
  # Issue the request again.
  users.fetch!
end
```


README.md
```diff
@@ -67,6 +67,13 @@ client = ZendeskAPI::Client.new do |config|
   # More information on obtaining OAuth access tokens can be found here:
   # https://developer.zendesk.com/api-reference/introduction/security-and-auth/#oauth-access-token
   config.access_token = "your OAuth access token"
+  # You can configure token refreshening by adding the OAuth client ID, secret and refresh token:
+  config.client_id = "your OAuth client id"
+  config.client_secret = "your OAuth client secret"
+  config.refresh_token = "your OAuth refresh token"
+  # Optionally you can set access and refresh token expiration:
+  config.access_token_expiration = 300 # time in seconds between 5 minutes and 2 days (300 and 172800)
+  config.refresh_token_expiration = 605800 # time in seconds between 7 and 90 days (605800 and 7776000)
```
bquorning: Should these be below the "Optional" header on line 78?

lib/zendesk_api/token_refresher.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      response = connection.post "/oauth/tokens" do |req|
+        req.body = {
+          grant_type: "refresh_token",
+          refresh_token: @config.refresh_token,
+          client_id: @config.client_id,
+          client_secret: @config.client_secret
+        }.tap do |params|
+          params.merge!(expires_in: @config.access_token_expiration) if @config.access_token_expiration
+          params.merge!(refresh_token_expires_in: @config.refresh_token_expiration) if @config.refresh_token_expiration
```
bquorning: merge! is not necessary:
```suggestion
          params[:expires_in] = @config.access_token_expiration) if @config.access_token_expiration
          params[:refresh_token_expires_in] = @config.refresh_token_expiration) if @config.refresh_token_expiration
```

peterkhuu:
Hi @zendesk-mradmacher, could I ask if this PR is something that is still being worked on? With the deprecation of API tokens in favour of access tokens, I think this refresh flow is something that is needed. If it's something that the library can support already that would be great since I think it'll mean other developers won't need to build it again 

https://developer.zendesk.com/documentation/api-basics/authentication/oauth-migration/

PR #595, opened 2025-10-12
Adding notifications support for various lifecycle conditions

cryptomail:
Allows applications to configure any instrumentation backend that responds to `#instrument(event_name, payload)`, e.g. `ActiveSupport::Notifications`

## Usage

```ruby
ZendeskAPI.configure do |config|
  config.instrumentation = ActiveSupport::Notifications
end
```
### `zendesk.request`
- When: All requests (emitted regardless of status code)
- Payload: `duration` (ms), `endpoint`, `method`, `status`

### `zendesk.rate_limit`
- When: Status < 500 AND rate limit headers present
- Not emitted for: 5xx errors, missing headers
- Payload: `endpoint`, `status`, `remaining`, `limit`, `reset`
- Note: Includes successful requests and 4xx errors when rate limit headers are present

### `zendesk.cache_hit`
- When: `304 Not Modified` served from cache
- Requires: Cache enabled, matching ETag
- Payload: `endpoint`, `status` (304)

### `zendesk.cache_miss`
- When: `200 OK` with ETag cached for future requests
- Requires: Cache enabled, ETag header present
- Payload: `endpoint`, `status` (200)

### `zendesk.retry`
- When: Before retry sleep (`429`, `503`, or exception)
- Requires: Retry enabled; exceptions need `retry_on_exception: true`
- Payload: `attempt` (starts at 1), `endpoint`, `method`, `delay` (seconds), `reason`
- Reasons: `"rate_limited"` (429), `"server_error"` (503), `"exception"` (network failures)

## Key assumptions/decisions
- Users must supply their own compatible `config.instrumentation`, this lib does not pull in `activesupport`
- `duration` reports round-trip time including network call and response processing middleware
- Rate limit events only for status < `500` (client/rate limit errors)
- Request events emitted for all requests (not filtered by status)
- Instrumentation failures logged at debug level or silently swallowed

### DIFF
--- a/lib/zendesk_api/middleware/request/etag_cache.rb
+++ b/lib/zendesk_api/middleware/request/etag_cache.rb
@@ -9,6 +9,7 @@ module ZendeskAPI
       class EtagCache < Faraday::Middleware
         def initialize(app, options = {})
           @app = app
+          @instrumentation = options[:instrumentation] if options[:instrumentation].respond_to?(:instrument)
           @cache = options[:cache] ||
             raise("need :cache option e.g. ActiveSupport::Cache::MemoryStore.new")
           @cache_key_prefix = options.fetch(:cache_key_prefix, :faraday_etags)
@@ -41,8 +42,30 @@ module ZendeskAPI
                 content_length: cached[:response_headers][:content_length],
                 content_encoding: cached[:response_headers][:content_encoding]
               )
+              if @instrumentation
+                begin
+                  @instrumentation.instrument("zendesk.cache_hit",
+                    {
+                      endpoint: env[:url]&.path,
+                      status: env[:status]
+                    })
+                rescue
+                  # Swallow instrumentation errors to maintain cache behavior
+                end
+              end
             elsif env[:status] == 200 && env[:response_headers]["Etag"] # modified and cacheable
               @cache.write(cache_key(env), env.to_hash)
+              if @instrumentation
+                begin
+                  @instrumentation.instrument("zendesk.cache_miss",
+                    {
+                      endpoint: env[:url]&.path,
+                      status: env[:status]
+                    })
+                rescue
+                  # Swallow instrumentation errors to maintain cache behavior
+                end
+              end
             end
           end
         end
--- a/lib/zendesk_api/middleware/request/retry.rb
+++ b/lib/zendesk_api/middleware/request/retry.rb
@@ -15,44 +15,79 @@ module ZendeskAPI
           @logger = options[:logger]
           @error_codes = (options.key?(:retry_codes) && options[:retry_codes]) ? options[:retry_codes] : DEFAULT_ERROR_CODES
           @retry_on_exception = (options.key?(:retry_on_exception) && options[:retry_on_exception]) ? options[:retry_on_exception] : false
+          @instrumentation = options[:instrumentation]
         end
 
         def call(env)
+          # Duplicate env for retries but keep attempt counter persistent
           original_env = env.dup
+          original_env[:call_attempt] = (env[:call_attempt] || 0)
+
           exception_happened = false
+          response = nil
+
           if @retry_on_exception
             begin
               response = @app.call(env)
-            rescue => e
+            rescue => ex
               exception_happened = true
+              exception = ex
             end
           else
+            # Allow exceptions to propagate normally when not retrying
             response = @app.call(env)
           end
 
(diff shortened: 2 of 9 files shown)


lib/zendesk_api/middleware/request/retry.rb
```diff
@@ -14,10 +14,16 @@ def initialize(app, options = {})
...
+          @instrumentation = options[:instrumentation]
         end
 
         def call(env)
           original_env = env.dup
+          if original_env[:call_attempt]
+            original_env[:call_attempt] += 1
+          else
+            original_env[:call_attempt] = 1
+          end
```
razumau: Can we have a test for values of these counts?

lib/zendesk_api/middleware/response/zendesk_request_event.rb
```diff
@@ -0,0 +1,41 @@
...
+                                           status: response_env[:status] })
+              if response_env[:status] < 500
+                instrumentation.instrument("zendesk.rate_limit",
+                                           {
+                                             endpoint: response_env[:url].path,
+                                             status: response_env[:status],
+                                             threshold: response_env[:response_headers] ? response_env[:response_headers][:x_rate_limit_remaining] : nil,
+                                             limit: response_env[:response_headers] ? response_env[:response_headers][:x_rate_limit] : nil,
+                                             reset: response_env[:response_headers] ? response_env[:response_headers][:x_rate_limit_reset] : nil
+                                           })
```
razumau: `threshold`, `limit`, and `reset` values don’t have any tests.

cryptomail: included in tests

lib/zendesk_api/middleware/request/etag_cache.rb
```diff
@@ -1,4 +1,5 @@
 require "faraday/middleware"
+require "active_support/notifications"
```
Thomascountz: We should be able to remove this since we're passing in `options[:instrumentation]` and don't use the module directly.

lib/zendesk_api/middleware/request/retry.rb
```diff
@@ -15,44 +15,74 @@ def initialize(app, options = {})
...
+            }
+          )
+        end
 
-            @app.call(original_env)
-          else
-            response
+        def sleep_with_logging(seconds_left)
+          seconds_left.times do |i|
+            sleep 1 if seconds_left > 0
```
Thomascountz: We can remove this redundant conditional. If we enter inside this loop `seconds_left` times, `seconds_left > 0` must always be `true`.

spec/core/middleware/request/retry_spec.rb
```diff
@@ -116,4 +116,76 @@ def runtime
...
+
+      event = retry_events.first[:payload]
+      expect(event[:attempt]).to eq(1)
+      expect(event[:endpoint]).to eq("/api/v2/instrumented")
+      expect(event[:method]).to eq(:get)
+      expect(event[:reason]).to eq("rate_limited")
+      expect(event[:delay]).to be >= 0
+    end
+
+    xit "instruments consecutive retry attempts" do
```
razumau: Should this and the following test be skipped?

lib/zendesk_api/middleware/response/zendesk_request_event.rb
```diff
@@ -0,0 +1,59 @@
...
+          @instrumentation = options[:instrumentation]
+          @logger = options[:logger]
+        end
+
+        def call(env)
+          start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          @app.call(env).on_complete do |response_env|
+            end_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+            duration = (end_time - start_time) * 1000.0
+            if @instrumentation
```
razumau: Nit: this method has a lot of nested ifs. Should some of these layers become something like `next unless @instrumentation`?

spec/core/middleware/request/retry_spec.rb
Thomascountz: I meant to leave a comment. The current behavior is that "retries" happen only once. If we want to change it, I think it should be in a follow up PR. I changed the implementation from `while` to `if` to avoid an infinite loop and to keep the behavior the same.

razumau (approved):

PR #597, opened 2025-10-16
rename brand in live specs

razumau:


cryptomail (approved):
I'm for it if it works

PR #598, opened 2025-10-17
adding new field mapping custom_field_symbol associative array, initi…

cryptomail:
…alized from global meta data garnered from account context

### DIFF
--- a/lib/zendesk_api/resources.rb
+++ b/lib/zendesk_api/resources.rb
@@ -464,6 +464,55 @@ module ZendeskAPI
     extend UpdateMany
     extend DestroyMany
 
+    # Proxy to trap array operator usage on custom_field_symbol
+    class CustomFieldSymbolProxy
+      def initialize(ticket)
+        @ticket = ticket
+        @field_array = @ticket.custom_fields || []
+      end
+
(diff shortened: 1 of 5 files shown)


lib/zendesk_api/resources.rb
```diff
@@ -464,6 +464,60 @@ class Ticket < Resource
...
+    end
+
+    def custom_field_symbol
+      @custom_field_symbol ||= CustomFieldSymbolProxy.new(self, @custom_field_symbol)
+    end
+
+    def custom_field_symbol=(val)
+      @custom_field_symbol = val
+      @custom_field_symbol_proxy = CustomFieldSymbolProxy.new(self, @custom_field_symbol)
+    end
```
Thomascountz: `custom_field_symbol=` overwrites `@custom_field_symbol` with the raw array and caches a proxy in `@custom_field_symbol_proxy`, but the getter still memoizes on `@custom_field_symbol`. After the first assignment (or any mass-assignment that sets this ivar) subsequent reads return the plain array and `ticket.custom_field_symbol['foo']` will raise a `TypeError` when trying to index an array with a string.

PR #617, opened 2026-07-22
PLATOPS-1274: Guard Setting#initialize against non-hash top-level settings

snbloch:
## What

`ZendeskAPI::Setting#initialize` assumed every top-level setting nests its value in a Hash and unconditionally ran:

```ruby
attributes.merge!(attributes.delete(@on))
```

A top-level **array** setting makes `attributes.delete(@on)` return an Array, and `merge!(Array)` raises:

```
TypeError: no implicit conversion of Array into Hash
```

This change only merges when the value under `@on` is a Hash; scalar/array settings are left in place under their own key.

## Why

Classic's `/api/v2/users/me/settings` gained a top-level `agent_home_pinned_views` array ([zendesk/zendesk#86115](https://github.com/zendesk/zendesk/pull/86115)), gated on the `has_agent_home?` account setting. Since `ZendeskAPI::User` declares `has_many Setting`, `user.settings.fetch!` instantiates one `Setting` per top-level key and crashes on the array. This is breaking **Pandora V1 account cleanup** (`update_owner_onboarding`).

This is a latent gem bug, not a defect in the Classic PR. `shared_views_order` is the same top-level-array shape and would crash identically; it has only been dodging the bug because it's gated behind the `user_level_shared_view_ordering` arturo, which isn't enabled on the accounts Pandora cleans.

## Write path is safe

`Collection#_save` only PUTs items where `changed?` is true, and an untouched non-hash setting stays `changed? == false`, so bulk `settings.save!` skips it. Namespaced Hash settings (`lotus`, `admin_center`) are unchanged — the merge guard uses the same string-key lookup that `@on` is derived from, so the hoisting behavior is identical to before.

## Tests

- New `spec/core/resources/setting_spec.rb`: namespaced Hash setting still hoists; top-level array setting no longer raises, stays accessible, and reports `changed? == false`; top-level scalar setting is handled too.
- Full core suite green (451 examples, 0 failures); `standard` lint clean.

## Follow-up (not in this PR)

Bump this gem in consumers so they pick up the guard: `pandora_v2`, `pandora`, `zendesk_api_tests`, `zendesk_browser_tests`.

## References

- Jira: https://zendesk.atlassian.net/browse/PLATOPS-1274

🤖 Generated with [Claude Code](https://claude.com/claude-code)


snbloch:
## CI note: the red check is the pre-existing flaky `Spec live` job, not this change

Every job that exercises this diff is green:

- **Core specs** on all 5 Rubies (3.2, 3.3, 3.4, 4.0, jruby-10.0) — `bundle exec rake`, which runs `spec/core/**` including the new `setting_spec.rb`. ✅
- **Lint** (`standard`) — ✅

The only failing job is **Spec live**, and all three failures are live-CI-account state, unrelated to this PR:

1. `ZendeskAPI::Voice::PhoneNumber creation should have an id` — `RecordInvalid: "You don't have any credit left to purchase a phone number."` The shared CI account is out of Voice credit.
2. `ZendeskAPI::Voice::PhoneNumber creation should be findable` — cascades from #1 (the create failed, so there's nothing to find).
3. `ZendeskAPI::OrganizationField deletion should be destroyable` — `undefined method 'active?' for nil`, i.e. the record wasn't in the expected state on the account.

These specs hit a real Zendesk account over the network via `spec/macros/resource_macros.rb` with no VCR cassette (the job `rm -rf`s cassettes first), so their outcome tracks that account's mutable state (billing balance, leftover fixtures), not the code. This diff only touches `ZendeskAPI::Setting#initialize` and goes nowhere near Voice, OrganizationField, or the create/delete macros. The `LocalJumpError: unexpected return` in the `after(:context)` hook is a separate pre-existing bug in the macro teardown (`resource_macros.rb:41`) that only surfaces once an earlier create fails.

No action needed on the diff. Flagging in case `Spec live` is a required check — if so it'll need a maintainer to top up the CI account's Voice credit / reset its org fields (or confirm the job is non-blocking), as that's outside what this PR can control.

codealchemy (approved):
👍

batizhevsky (approved):

HeyNonster (approved):

